PR #48, opened 2018-08-27
Remove legacy linting rules

ccnixon:
**What does this PR do?**
Removes the legacy linting rules and replaces them with a general rule to just flag any non-es5 syntax. This will allow us to use Standard as our code formatter while also protecting against accidentally shipping code that potentially breaks our supported browsers.

**Link to CC ticket**

**Helpful Docs**

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,3 @@
 {
-  "extends": "@segment/eslint-config/browser/legacy"
+  "ecmaVerion": 5
 }


f2prateek:
for refernce, here's how we integrated prettier into analytics.js-core https://github.com/segmentio/analytics.js-core/pull/89

ccnixon:
Thanks @f2prateek - I like that. @fathyb - from what you can tell, does the legacy eslint config help protect us more than just setting the `ecmaVersion` to 3?

ccnixon:
It's tough to find compatibility tables for ie8-10

fathyb:
@ccnixon 
> does the legacy eslint config help protect us more than just setting the ecmaVersion to 3?

Yes, mainly to cope with IE8 quircks: 
```js
  "rules": {
    "comma-dangle": ["error", "never"],
    "dot-notation": ["error", { "allowKeywords": false }],
    "no-catch-shadow": "error",
    "no-restricted-globals": ["error", "JSON"], // IE7
    "quote-props": ["error", "as-needed", { "keywords": true }],
    "radix": ["error", "always"]
  }
```

- `comma-dangle` prevents you from using trailing commas in object literals (which are legal per ES3 but IE8 doesn't support it)
- `dot-notation` will prevent you from using a keyword (ie. `default`) in a member expression
- `no-catch-shadow` prevent a bug in IE8 when a `catch` arguments shadows a variable in the outer scope
- `no-restricted-globals` to prevent using `JSON` (and use something like `require('json')` instead)
- `quote-props` same as `dot-notation` but for object literals
- `radix` because the behaviour of `parseInt` is inconsistent between ES3 and ES5

Peripheral1994:
FWIW, we've formally deprecated support for IE8 and earlier, though I'm 👍 for continuing to informally support it if it isn't a blocker for us long-term. :)

Peripheral1994:
e.g. We use Array.isArray in several destinations now on the clientside as a result, IIRC.

ccnixon:
Great, thanks @fathyb!

PR #51, opened 2018-09-06
Fix linting

ccnixon:
**What does this PR do?**
- updates `.eslintrc` to use the segment legacy browser config as well as prettier
- adds a pre-commit hook to lint and format staged files
- runs the formatter over the whole repo to standardize

All tests are passing, no issues with the formatting.

**Link to CC ticket**

**Helpful Docs**


f2prateek:
😍

integrations/friendbuy/test/index.test.js
```diff
@@ -63,7 +63,7 @@ describe('FriendBuy', function() {
     });
   });
 
-  describe('loading', function() {
```
f2prateek: intentional?

ccnixon: Yeah I'm having issues with Friendbuy timing out loading their script in CI. It works fine locally. I'm skipping it for now and am going to test it separately in a new PR so that we don't need to run ALL tests each time.

PR #59, opened 2018-09-10
FB Pixel: Custom Content Types and Automatic Configuration

gpsamson:
**What does this PR do?**
- Adds a setting to disable [Automatic Configuration](https://developers.facebook.com/docs/facebook-pixel/api-reference#automatic-configuration).
- Adds a mapping for Content Types ([FB Pixel Settings](https://app.segment.com/partner-portal/integration/facebook-pixel/settings)). For backwards compatibility we are falling back to `product` and `product_group`.

Negative.

TODO in staging.

Local `automaticConfiguration` testing was verified in dev tools.

Local testing where a custom content type is passed to FB:
screen shot 2018-09-12 at 7 10 37 pm
The category "Games" was mapped to "vehicles" in this example.

Two new [settings](https://app.segment.com/partner-portal/integration/facebook-pixel/settings) were added, but migrations weren't needed.

[PLATFORM-2538](https://segment.atlassian.net/browse/PLATFORM-2538)
[PLATFORM-2310](https://segment.atlassian.net/browse/PLATFORM-2310)

**Link to CC ticket**
https://segment.atlassian.net/browse/CC-1573

Personal FB Pixel account

**Helpful Docs**
https://paper.dropbox.com/folder/show/Facebook-Pixel-e.1gg8YzoPEhbTkrhvQwJ2zz409feT37UwVJoNokfnCOndQwCZywN3

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -24,6 +24,7 @@ var FacebookPixel = module.exports = integration('Facebook Pixel')
   .option('valueIdentifier', 'value')
   .option('initWithExistingTraits', false)
   .option('traverse', false)
+  .option('disableAutoConfig', false)
   .mapping('standardEvents')
   .mapping('legacyEvents')
   .tag('<script src="//connect.facebook.net/en_US/fbevents.js">');
@@ -52,6 +53,9 @@ FacebookPixel.prototype.initialize = function() {
   window.fbq.version = '2.0';
   window.fbq.queue = [];
   this.load(this.ready);
+  if (this.options.disableAutoConfig) {
+    window.fbq('set', 'autoConfig', false, this.options.pixelId);
+  }
   if (this.options.initWithExistingTraits) {
     var traits = formatTraits(this.analytics);
     window.fbq('init', this.options.pixelId, traits);
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -191,7 +191,7 @@ FacebookPixel.prototype.productListViewed = function(track) {
 
   window.fbq('track', 'ViewContent', {
     content_ids: contentIds,
-    content_type: contentType
+    content_type: track.proxy('properties.contentType') || contentType, 
   });
 
   // fall through for mapped legacy conversions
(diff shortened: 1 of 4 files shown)


jlee9595:
having a setting to explicitly 'disable autoconfig' sounds unnecessarily confusing to me. would it be possible to just have the setting be 'autoconfig', which can be disabled or enabled? i feel that would eliminate some redundancy there and make things more straightforward.

gpsamson:
I agree - "Automatic Configuration" makes more sense for naming. This would mean the default value would be `true` and we'd need to backfill. If this checks out with you it works for me.

jlee9595:
Yeah that sounds perfect and I believe that's how we've handled these kinds of situations in the past

jlee9595:
Looking at the FB docs looks like there's also the option to send multiple content_types. Is that something that makes sense for us to support now or does it make more sense to punt on it till a customer needs it?

integrations/facebook-pixel/lib/index.js
```diff
@@ -52,6 +54,9 @@ FacebookPixel.prototype.initialize = function() {
   window.fbq.version = '2.0';
   window.fbq.queue = [];
   this.load(this.ready);
+  if (!this.options.automaticConfiguration) {
```
gpsamson: This conditional is relevant to [PLATFORM-2310](https://segment.atlassian.net/browse/PLATFORM-2310).

All changes below this belong to [PLATFORM-2538](https://segment.atlassian.net/browse/PLATFORM-2538)

integrations/facebook-pixel/test/index.test.js
```diff
@@ -86,6 +90,20 @@ describe('Facebook Pixel', function() {
         analytics.assert(window.fbq instanceof Function);
       });
 
+      before(function() {
```
ccnixon: I would omit these `before` and `after` blocks and just explicitly toggle this option before and after the test code runs. So like this:

```
      it('should call set autoConfig if option disableAutoConfig is enabled', function () {
options. automaticConfiguration = false
        analytics.stub(window, 'fbq');
        analytics.initialize();
        analytics.called(window.fbq, 'set', 'autoConfig', false, options.pixelId);
options.automaticConfiguration = true;
      });
```

ccnixon: Unless you think there would be a need for future tests, in which case, I would use `beforeEach` and `afterEach`.

gpsamson: Ahh, doing it in the test code doesn't work since a.js is initialized before-hand.

PR #62, opened 2018-09-13
Add PII filtering

ccnixon:
**What does this PR do?**
Addresses: https://segment.atlassian.net/browse/PLATFORM-3100 by adding filters to check for PII data. The keys we check for are pulled from here: https://developers.facebook.com/docs/facebook-pixel/pixel-with-ads/conversion-tracking/. 

PR also introduces a new setting called `whitelistPiiProperties`. This will allow user's to optionally choose to bypass FB's PII policy and whitelist properties.

No

**Link to CC ticket**
https://segment.atlassian.net/browse/CC-1619

**Helpful Docs**

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -17,7 +17,7 @@ var Track = require('segmentio-facade').Track;
  * Expose `Facebook Pixel`.
  */
 
-var FacebookPixel = module.exports = integration('Facebook Pixel')
+var FacebookPixel = (module.exports = integration('Facebook Pixel')
   .global('fbq')
   .option('pixelId', '')
   .option('agent', 'seg')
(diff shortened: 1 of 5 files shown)


integrations/facebook-pixel/lib/index.js
```diff
@@ -94,46 +95,74 @@ FacebookPixel.prototype.page = function() {
...
+
+      var dateFields = [
+        'checkinDate',
+        'checkoutDate',
+        'departingArrivalDate',
+        'departingDepartureDate',
+        'returningArrivalDate',
+        'returningDepartureDate',
+        'travelEnd',
+        'travelStart'
```
jlee9595: maybe we can define these constants and the pii constants at the top of the file so only the filtering logic lives inside the `foldl`?

PR #76, opened 2018-10-10
Prevent SFDMP from throwing internal errors

brennan:
**What does this PR do?**
This PR prevents SFDMP from throwing errors when an unmapped `track` event is invoked while the `trackEventNames` setting is enabled.

n/a

n/a

Should reduce `track`-related errors to 0 - https://segment.datadoghq.com/dash/807059/analyticsjs-integration-dashboard?live=true&page=0&is_auto=false&from_ts=1539186716844&to_ts=1539201116844&tile_size=m&tpl_var_integration_name=salesforce_dmp

n/a

n/a

n/a

**Link to CC ticket**

n/a

**Helpful Docs**
n/a


calvinfo:
This looks reasonable, can we add test coverage to catch this case? There should be some accompanying test here which fails without the modification

brennan:
@calvinfo Will do!

brennan:
@calvinfo Having trouble using our test suite to validate that a method doesn't throw an error. We can come back and add a test for this resolved edge case later; otherwise, we may not be able to deploy the fix today. Let me know what you think!

integrations/salesforce-dmp/test/index.test.js
```diff
@@ -188,6 +189,27 @@ describe('Salesforce DMP V2', function () {
         analytics.track('event')
         analytics.assert(window.kruxDataLayer['event_name'] === 'do not override')
       })
+
+      // previously SFDMP would throw an error if a customer invoked an 
+      // unmapped `track` event while `sendEventNames` setting was enabled
+      // this edge case has been resolved, but we test for it anyway
+      it.only('throws an error if `sendEventNames` setting is enabled but no track events are whitelisted', function () {
```
ucarion: I think the `only` thing is accidentally left over?

integrations/salesforce-dmp/lib/index.js
```diff
@@ -117,6 +117,8 @@ SalesforceDMP.prototype.identifyV2 = SalesforceDMP.prototype.trackV2 = function
         }
         if (i === this.options.trackFireEvents.length - 1) return
       }
+    } else {
+      return
```
ucarion: This seems like a pretty big change? Do we have tests for this early return?

integrations/salesforce-dmp/lib/index.js
```diff
@@ -137,22 +139,24 @@ SalesforceDMP.prototype.identifyV2 = SalesforceDMP.prototype.trackV2 = function
...
   }
 
-  if (this.options.sendEventNames && !window.kruxDataLayer['event_name']) {
-    var name
-    if (type === 'track') {
-      name = msg.event()
-    } else {
-      if (this.options.sendIdentifyAsEventName) {
-        name = type
+  if (window.kruxDataLayer) {
```
ucarion: The diff GitHub is showing here is just from indenting the code because of this `if`, right?

integrations/salesforce-dmp/test/index.test.js
```diff
@@ -188,6 +189,27 @@ describe('Salesforce DMP V2', function () {
...
+        var edgeCaseOptions = {
+          confId: 'rw1to29bb',
```
ucarion: Is this value sensitive?

integrations/salesforce-dmp/lib/index.js
ucarion: Hmm, actually it seems like this early return is the meat of your fix. Maybe we could put a comment explaining why it needs to be there / what the consequences of leaving it out are?

PR #85, opened 2018-10-24
Rockerbox: adding userId to identify calls

patrickotoole:
**What does this PR do?**
- Adds format to identify call before sending to Rockerbox
- Adds customer_id overridden as userId if userId is present

No

This populates a new set of fields that we can use to override the default customer identifier in our system. This is a feature supported by our product currently but now passes userId and anonymousId from segment to our product.

This will enable our customers to use the segment userID in place of the Rockerbox customer_id and simplify setup by relying on segments userID

NA

NA

NA

**Link to CC ticket**
NA

**Helpful Docs**


skarim:
@calvinfo we've updated the PR, but it appears that some of the circleci tests are pending/not running. can you please help here?

PR #96, opened 2018-12-12
Add Pinterest Enhanced Match

dja:
**What does this PR do?**
Adds Pinterest Enhanced Match, if optionally specified. See: https://developers.pinterest.com/docs/ad-tools/enhanced-match/

Not sure that the tests are perfectly done, but would love your help on that!

Improves attribution considering Safari Intelligent Tracking Protocol

This does require an extra option to be enabled by users.

**Link to CC ticket**

**Helpful Docs**
https://developers.pinterest.com/docs/ad-tools/enhanced-match/


srthurman:
Hi @dja - Sara from Segment here.
Thanks for submitting this PR. Currently we don't have a timeline we can commit to regarding when we' may be able to merge this update. We are working on ways to make contributions from dedicated customers like yourself faster and easier in the future though!

platedbay:
Hi folks, just wanted to add a +1 here from us as this is definitely needed on our end. Was actually looking into opening a PR just for this and it's good to see it here already. Please let us know whatever we can do to try to get this through.

mericsson:
@CarlosMecha we can close this PR now right because we deployed it?

PR #97, opened 2018-12-13
Adds exit link tracking support for specific events

gpsamson:
**What does this PR do?**
Adds support for a new settings which lists events which Segment Events we should track as exit links. 

Negative

screen shot 2018-12-13 at 10 47 48 am

Will update with SS PR.

https://segment.atlassian.net/browse/FCD-70

**Link to CC ticket**
TODO

**Helpful Docs**

### DIFF
--- a/integrations/adobe-analytics/lib/index.js
+++ b/integrations/adobe-analytics/lib/index.js
@@ -83,6 +83,7 @@ AdobeAnalytics
   .option('preferVisitorId', false)
   .option('heartbeatTrackingServerUrl', '')
   .option('ssl', false)
+  .option('exitLinkEvents', [])
 
   .sOption('visitorID')
   .sOption('channel')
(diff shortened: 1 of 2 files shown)


jlee9595:
Code looks great but wanted to flag: if I'm understanding exit links correctly they'd never be used in the context of our specced ecommerce events. Therefore customers would only be using them for their custom events that they're mapping via the existing `events` setting (that is to be upgraded by Ladan and I). If this is the case then I think it would make more sense to do a checkbox in the event mapping mixed component rather than another list of events.

I'm far from positive that these wouldn't be used for specced ecommerce events but I just wanted to make sure we'd considered

gpsamson:
@jlee9595 That makes sense, and you're right. More often than not our ecommerce events won't be tracked as exit links. But per an offline discussion with Sara we decided to introduce a new setting for a couple reasons:
- To avoid having to alter an existing setting and having to backfill for existing users
- Customers tend to have anomalous implementations and could possibly want to track ecomm events as exit links.

Open to have a discussion about this in Slack, though!

jlee9595:
Got it. If it were up to me I'd track back and ask the requesting customer(s) (I think you mentioned there weren't many) whether leaving out eccom events would work for them before jumping to the conclusion that customers _might_ want to use an anomalous implementation. If they'd be fine with leaving out ecom events then we could then do this the 'right way' and avoid the extra setting. Backfilling is a pretty trivial thing to do and we'll already be doing it for the planned upgrade to the `events` setting.

That being said I think you know everything that I do at this point so I'll leave it up to you to make the decision.

gpsamson:
Closing this PR as we decided to punt on this.

PR #104, opened 2019-02-19
enabled enhanced match by intercepting the identify call

durgani:
**What does this PR do?**
Enables Pinterest Enhanced Match feature by sending an email address to the pinterest load function

Since this requires re-loading the Pinterest library, not sure of the unintended consequences.  Need to isolate this to a given customer;s workspace for testing.  

No

Ritual, a Segment customer, needs enhanced match to improve their conversion tracking on Pinterest especially when the Pinterest 3rd party cookie is not available.  This requires passing the user's email to Pinterest during an identify call.  

N/A

**Link to CC ticket**

**Helpful Docs**
https://help.pinterest.com/en/business/article/track-conversions-with-pinterest-tag

### DIFF
--- a/integrations/pinterest-tag/lib/index.js
+++ b/integrations/pinterest-tag/lib/index.js
@@ -36,6 +36,16 @@ Pinterest.prototype.loaded = function () {
   return !!(window.pintrk && window.pintrk.queue && window.pintrk.queue.push !== Array.prototype.push)
 }
 
+Pinterest.prototype.identify = function (identify) {
+  // If we have an email then enable Enhanced Match feature by reloading the Pinterest library. 
+  // TODO: We may want to add a toggle in the Pinterest integration UI as a configuration to enable enhanced match
+  if (identify.email()) {
+    pintrk('load', this.options.tid, {
+      em: identify.email(),
+    })
+  }
+}
+
 Pinterest.prototype.page = function (page) {
   // If we have a category, the use ViewCategory. Otherwise, use a normal PageVisit.
   if (page.category()) {


f2prateek:
From the pinterest docs, it sounds like we want to do the initial load with an email, rather than reload the library. Is there anything in their docs which mentions how to handle reloads?

integrations/pinterest-tag/test/index.test.js
```diff
@@ -63,4 +63,28 @@ describe('Pinterest', function () {
...
+      })
+
+      it('should not fire the Pinterest pixel tag', function () {
+        analytics.identify()
+
+      })
+
+      it('should push Segment email to Pinterest Enhanced Match', function () {
+        analytics.identify('123', { email: 'prakash@segment.com'})
+        analytics.called(window.Pinterest)
```
f2prateek: I think this should be `analytics.called(pinterest.load)`. We might also want to validate the arguments (i.e. the email)

integrations/pinterest-tag/test/index.test.js
```diff
@@ -63,4 +63,28 @@ describe('Pinterest', function () {
...
+    })
+
+    describe('#identify', function () {
+      beforeEach(function () {
+        analytics.stub(window, 'Pinterest')
+      })
+
+      it('should not fire the Pinterest pixel tag', function () {
+        analytics.identify()
+
```
f2prateek: `analytics.didNotCall(pinterest.load)`

integrations/pinterest-tag/test/index.test.js
```diff
@@ -63,4 +63,27 @@ describe('Pinterest', function () {
...
+        analytics.stub(window, 'Pinterest')
+      })
+
+      it('should not fire the Pinterest pixel tag', function () {
+        analytics.identify()
+	analytics.didNotCall(pinterest.load)
+      })
+      it('should push Segment email to Pinterest Enhanced Match', function () {
+        analytics.identify('123', { email: 'prakash@segment.com'})
+        analytics.called(pinterest.load)
```
f2prateek: let's change this to `analytics.called(pinterest.load, 2620795819800, {em: "prakash@segment.com"})`

PR #106, opened 2019-02-28
Move Woopra integration

CarlosMecha:


integrations/woopra/lib/index.js
```diff
@@ -0,0 +1,133 @@
...
+  .tag('<script src="//static.woopra.com/js/w.js">');
+
+/**
+ * Initialize.
+ *
+ * http://www.woopra.com/docs/setup/javascript-tracking/
+ */
+
+Woopra.prototype.initialize = function() {
+  /* eslint-disable */
```
maxence-charriere: I guess this is supposed to be normal but if we could have a comment about the where it come from and why.

integrations/woopra/lib/index.js
```diff
@@ -0,0 +1,133 @@
...
+  (function(){var i, s, z, w = window, d = document, a = arguments, q = 'script', f = ['config', 'track', 'identify', 'visit', 'push', 'call'], c = function(){var i, self = this; self._e = []; for (i = 0; i < f.length; i++){(function(f){self[f] = function(){self._e.push([f].concat(Array.prototype.slice.call(arguments, 0))); return self; }; })(f[i]); } }; w._w = w._w || {}; for (i = 0; i < a.length; i++){ w._w[a[i]] = w[a[i]] = w[a[i]] || new c(); } })('woopra');
```
maxence-charriere: I guess this is supposed to be normal but if we could have a comment about the where it come from and why.

CarlosMecha: Are you referring to the snippet? or the integration itself? If it's the later, we have it :) Check the [commit](https://github.com/segmentio/analytics.js-integrations/pull/106/commits/a724dd909f7d2da8a76d4efb94a81efe00a2feaf) and the [original repo](https://github.com/segment-integrations/analytics.js-integration-woopra).

PR #107, opened 2019-02-28
Woopra context

maxence-charriere:
**What does this PR do?**
- Adds context to Woopra event.
- Adds no context override test in Woopra tests.
- Adapt tests to work with latest Woopra changes.

no

No

NA

no

https://segment.atlassian.net/browse/FCD-405

**Link to CC ticket**

**Helpful Docs**
NA

### DIFF
--- a/integrations/woopra/lib/index.js
+++ b/integrations/woopra/lib/index.js
@@ -17,7 +17,7 @@ var time = require('unix-time');
  * Expose `Woopra` integration.
  */
 
-var Woopra = module.exports = integration('Woopra')
+var Woopra = (module.exports = integration('Woopra')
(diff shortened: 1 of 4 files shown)


integrations/woopra/lib/index.js
```diff
@@ -107,7 +107,12 @@ Woopra.prototype.identify = function(identify) {
  */
 
 Woopra.prototype.track = function(track) {
-  window.woopra.track(track.event(), stringifyNested(track.properties()));
+  var properties = track.properties();
+  if (!properties.context) {
```
f2prateek: This should probably be behind a flag so customers can opt into this behaviour.

PR #120, opened 2019-03-04
Woopra set context

maxence-charriere:
**What does this PR do?**
This PR set Woopra event context properly.

No

See below.

From conversation with Woopra:

**Question**:

I tried it and put a log to ensure that the context is populated in the config.

screen shot 2019-03-04 at 12 22 00 pm

Tough, look like the context is not passed as a query parameter when I’m sending a track event right after the config call:
screen shot 2019-03-04 at 12 22 51 pm
Is it expected?

**Answer**:

Hi Maxence, 
Yes it makes perfect sense, we will be rolling out a new tracker script shortly that would pass it, we have our segment customers using the new tracker script already. So we should be fine.
Thank you.

This ensures parity with server side.

No

https://segment.atlassian.net/browse/FCD-405

**Link to CC ticket**

**Helpful Docs**

### DIFF
--- a/integrations/woopra/lib/index.js
+++ b/integrations/woopra/lib/index.js
@@ -43,7 +43,7 @@ var Woopra = (module.exports = integration('Woopra')
 
 Woopra.prototype.initialize = function() {
   /* eslint-disable */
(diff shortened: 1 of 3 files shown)


maxence-charriere:
Thanks for review.

PR #121, opened 2019-03-05
Prateek/enhanced match pinterest

ccnixon:
What does this PR do?
Enables Pinterest Enhanced Match feature by sending an email address to the pinterest load function

Since this requires re-loading the Pinterest library, not sure of the unintended consequences. Need to isolate this to a given customer;s workspace for testing.

Has this been tested end-to-end? Please provide screenshots on how the fix now populates in the end tool. If not, what was done to test?
No

A Segment customer needs enhanced match to improve their conversion tracking on Pinterest especially when the Pinterest 3rd party cookie is not available. This requires passing the user's email to Pinterest during an identify call.

Is there parity with the server-side/android/iOS integration (if applicable)?
N/A

Does this require a metadata change? If so, please link the PR from https://github.com/segmentio/destination-scripts.

What are the relevant tickets?

Link to CC ticket

List all the tests accounts you have used to make sure this change works

Helpful Docs
https://help.pinterest.com/en/business/article/track-conversions-with-pinterest-tag

### DIFF
--- a/integrations/pinterest-tag/lib/index.js
+++ b/integrations/pinterest-tag/lib/index.js
@@ -1,78 +1,80 @@
-'use strict'
+'use strict';
 
 /**
  * Module dependencies.
  */
 
-var integration = require('@segment/analytics.js-integration')
-var analyticsEvents = require('analytics-events')
+var integration = require('@segment/analytics.js-integration');
+var analyticsEvents = require('analytics-events');
 
 /**
  * Expose `Pinterest` integration.
  */
 
-var Pinterest = module.exports = integration('Pinterest Tag')
+var Pinterest = (module.exports = integration('Pinterest Tag')
   .global('pintrk')
   .option('tid', '')
   .option('pinterestCustomProperties', [])
   .mapping('pinterestEventMapping')
-  .tag('<script src="https://s.pinimg.com/ct/core.js"></script>')
+  .tag('<script src="https://s.pinimg.com/ct/core.js"></script>'));
 
-Pinterest.prototype.initialize = function () {
+Pinterest.prototype.initialize = function() {
   // We require a Tag ID to run this integration.
-  if (!this.options.tid) return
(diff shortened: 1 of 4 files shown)


ccnixon:
Unfortunately the formatter ran automatically and got merged in with my commit :(

I added some comments inline in the code that explain the main changes.

integrations/pinterest-tag/lib/index.js
```diff
@@ -1,68 +1,80 @@
...
+  window.pintrk('load', this.options.tid);
+  window.pintrk('page'); // This is treated semantically different than our own page implementation.
 
-  this.createPropertyMapping()
-}
+  this.createPropertyMapping();
+};
 
-Pinterest.prototype.loaded = function () {
-  return !!(window.pintrk && window.pintrk.queue && window.pintrk.queue.push !== Array.prototype.push)
```
ccnixon: This was actually never resolving to `true`. The pinterest tag seems to load another tag so there's no way we can actually hook into that script load lifecycle and delay initializing until it's finished.

That being said it's not necessary to because we interface with their synchronous API (`pintrk()`) when triggering events so we don't need to wait until their full script loads to start tracking data.

integrations/pinterest-tag/lib/index.js
```diff
@@ -1,68 +1,80 @@
...
-}
+Pinterest.prototype.loaded = function() {
+  return !!(window.pintrk && window.pintrk.queue);
+};
 
-Pinterest.prototype.page = function (page) {
+Pinterest.prototype.identify = function(identify) {
```
ccnixon: This was added by @durgani. @durgani can you quickly re-confirm that Pinterest confirmed that it's ok the run the `load` action twice on the same page.

f2prateek: From reading Pinterest docs, I can confirm they say this is ok.

PR #122, opened 2019-03-07
[FB Pixel] Add deduping with server-side

jlee9595:
In advance of allowing customers to use the client-side pixel integration with the server-side component simultaneously, we need to add deduping functionality. This adds `messageId` in the `eventID` field for Pixel events to achieve that (the server-side will do the same).

In addition, at Pixel's recommendation we're changing `track` calls to `trackSingle` calls. The difference between the two is that `track` makes track calls for all registered pixel instances on the page whereas `trackSingle` only makes the track call for the specified pixel instance. This will not result in any functional change for the integration as the integration only registers one pixel instance but it prevents a scenario where customers have other native FB pixel instances loaded and our code triggers a track for that pixel.

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -93,6 +93,7 @@ FacebookPixel.prototype.page = function() {
  */
 
 FacebookPixel.prototype.track = function(track) {
+  var self = this;
   var event = track.event();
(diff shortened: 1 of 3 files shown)


jlee9595:
That extra parameter is required when you use `trackSingle` instead of `track` as we are now specifying that we're using a specific single pixel instance so we need to provide the id of that pixel.

PR #127, opened 2019-04-03
Add lerna and hoist dev dependencies

ccnixon:
**What does this PR do?**
Adds lerna to help improve our dependency graph mess. Also hoists dev dependencies across all integration packages to the root package.json file. This will make installs about 100x faster.

**This does not deploy anything to production.**

No

**Link to CC ticket**

**Helpful Docs**


ccnixon:
Confirmed that the `nielsen-dcr` tests are not passing locally on master so those are not related to this change.

PR #128, opened 2019-04-03
[DEST-503][CC-???] Doubleclick/cookie match

ccnixon:
Adds support for DBM cookie matching functionality.

- [DEST-503](https://segment.atlassian.net/browse/DEST-503)
- [CC-???]()


ccnixon:
Lots of force pushes here. Had to debug an issue with the tests in IE and was having trouble getting SauceLabs working locally 😢

integrations/doubleclick-floodlight/lib/index.js
```diff
@@ -28,6 +42,28 @@ var Floodlight = module.exports = integration('DoubleClick Floodlight')
...
+  };
+  // In the initialize method:
+  // Check if we should load the DoubleClick ID pixel (and only proceed if we haven't already done so).
+  if (this.options.getDoubleClickId && this.options.googleNetworkId) {
+    if (!cookie('doubleclick_id_ts')) {
+      // Load the doubleclick pixel.
+      this.load('doubleclick id', {
+        googleNetworkId: this.options.googleNetworkId,
+        segmentWriteKey: this.options.segmentWriteKey,
+        // TODO: handle userId being nulls/undefined.
```
ccnixon: @f2prateek - what would the concern here be if `userId` was undefined? Should this be something that's handled on Personas end?

f2prateek: possibly - also wasn't sure how loading the pixel would be handled with a null parameter

integrations/doubleclick-floodlight/lib/index.js
```diff
@@ -28,6 +41,18 @@ var Floodlight = module.exports = integration('DoubleClick Floodlight')
...
 
 Floodlight.prototype.initialize = function() {
+  // In the initialize method:
+  // Check if we should load the DoubleClick ID pixel (and only proceed if we haven't already done so).
+  if (this.options.getDoubleClickId && this.options.googleNetworkId) {
+    // Load the doubleclick pixel.
+    this.load('doubleclick id', {
+      googleNetworkId: this.options.googleNetworkId,
+      segmentWriteKey: this.options.segmentWriteKey,
+      // TODO: handle userId being nulls/undefined.
```
andreiko: Yeah, currently we end up with `user_id=null` in img src which is confusing for idsync-service. I'll filter this value out for now.

ccnixon: Hey @andreiko - should we only fire the pixel if `userId` exists?

andreiko: Can we just remove the `user_id` parameter from the pixel url in that case?

ccnixon: We can but is it not required to actually facilitate the match or can you do it with `anonymousId`?

andreiko: yeah, it would work with just the `anonymousId`

andreiko:
I noticed that the dev version of this integration that we installed on nightmarejs.org stopped working on July 05. My guess is that it might have been switched back to master. 

Are there any plans to merge this?

CarlosMecha:
Taking it from here. Thanks!

PR #132, opened 2019-06-05
Add support for pii blacklist setting

ccnixon:
**What does this PR do?**
Supports a user defined property blacklist setting. This change is a result of user's requesting that they define which of their event properties contain PII as our default list is too narrow.

No

Yes

N/A

https://segment.atlassian.net/browse/DEST-555

**Link to CC ticket**
https://segment.atlassian.net/browse/CC-4778

**Helpful Docs**

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -12,6 +12,7 @@ var camel = require('to-camel-case');
 var is = require('is');
 var dateformat = require('dateformat');
 var Track = require('segmentio-facade').Track;
+var sha256 = require('js-sha256');
 
 /**
  * Expose `Facebook Pixel`.
@@ -26,6 +27,7 @@ var FacebookPixel = (module.exports = integration('Facebook Pixel')
   .option('traverse', false)
   .option('automaticConfiguration', true)
   .option('whitelistPiiProperties', [])
+  .option('blacklistPiiProperties', [])
   .mapping('standardEvents')
   .mapping('legacyEvents')
   .mapping('contentTypes')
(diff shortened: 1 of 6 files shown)


integrations/facebook-pixel/lib/index.js
```diff
@@ -97,6 +99,7 @@ FacebookPixel.prototype.track = function(track) {
   var event = track.event();
   var revenue = formatRevenue(track.revenue());
   var whitelistPiiProperties = this.options.whitelistPiiProperties || [];
+  var customPiiProperties = this.options.blacklistPiiProperties || [];
```
mericsson: Any reason to call this `customPiiProperties` vs keeping it as `blacklistPiiProperties`? Confused me for a second.

integrations/facebook-pixel/lib/index.js
```diff
@@ -26,6 +27,7 @@ var FacebookPixel = (module.exports = integration('Facebook Pixel')
   .option('traverse', false)
   .option('automaticConfiguration', true)
   .option('whitelistPiiProperties', [])
+  .option('blacklistPiiProperties', [])
```
mericsson: Does this mean we also have to add UI to have this setting in the Destination UI?

integrations/facebook-pixel/lib/index.js
ccnixon: Yeah I chose it to differentiate between the `defaultPiiProperties` variable that exists in the `track` function where this setting is used. I figured custom/default was a better indication of what each was doing.

integrations/facebook-pixel/lib/index.js
ccnixon: Yeah it's a net new addition though so no need to do anything funky with a DB migration which is nice.

PR #139, opened 2019-06-18
[DEST-577][CC-4816] Bump Amplitude SDK to 5.2.2

CarlosMecha:
Decided to bump the integration version to 3 since is also mayor version for Amplitude.

- [DEST-577](https://segment.atlassian.net/browse/DEST-577)
- [CC-4816](https://segment.atlassian.net/browse/CC-4816)

Nope.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,17 @@ jobs:
           root: .
           paths: [.]
 
+  lint:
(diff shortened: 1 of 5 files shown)


CarlosMecha:
The linter requires to change quite a lot of code, I'm skipping that for now.

PR #137, opened 2019-06-13
Update monetate tracking

manbearshark:
**What does this PR do?**

Adds support for Product List Viewed eCommerce event and also changes the Monetate API called for Product Viewed, since we were using the incorrect one.

**Are there breaking changes in this PR?**

No.

**Has this been tested end-to-end? Please provide screenshots on how the fix now populates in the end tool. If not, what was done to test?**

Manual test...

**Any background context you want to provide?**

This was asked for by JUUL as part of their rollout for Monetate.  The Monetate team noticed that certain events were not populating as expected.

**Is there parity with the server-side/android/iOS integration (if applicable)?**

No.

**Does this require a metadata change? If so, please link the PR from https://github.com/segmentio/destination-scripts.**

No.

**What are the relevant tickets?**

**Link to CC ticket**

**List all the tests accounts you have used to make sure this change works**

**Helpful Docs**

PR #142, opened 2019-06-24
[DEST-757][CC-4858] Add product list viewed event to Monetate

CarlosMecha:
- Original PR: #137 
- [DEST-757](https://segment.atlassian.net/browse/DEST-757)
- [CC-4854](https://segment.atlassian.net/browse/CC-4854)

### DIFF
--- a/integrations/monetate/lib/index.js
+++ b/integrations/monetate/lib/index.js
@@ -84,8 +84,8 @@ Monetate.prototype.productListViewed = function(track) {
 
   each(products, function(product) {
     var track = new Track({ properties: product });
-    product = toProducts(track);
-    items.push(product);
+    var p = toProducts(track);
+    items.push(p);
   });
 
   push(this.options.events.productListViewed, items);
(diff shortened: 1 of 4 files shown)


integrations/monetate/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@segment/analytics.js-integration-monetate",
   "description": "The Monetate analytics.js integration.",
-  "version": "2.2.0",
```
ccnixon: If it's not a breaking change, should we just bump by a minor version?

CarlosMecha: Sadly is a breaking change. It changes how `product viewed` behaves :(

ccnixon: Oh it doesn't just add a new handler for `productListViewed`?

integrations/monetate/lib/index.js
```diff
@@ -40,7 +40,8 @@ Monetate.prototype.initialize = function() {
   if (this.options.retail) {
     this.options.events = {
       orderCompleted: 'addPurchaseRows',
-      productViewed: 'addProducts',
+      productViewed: 'addProductDetails',
```
ccnixon: Is this the breaking change?

CarlosMecha: Yep!

PR #131, opened 2019-05-24
Add content_type to FB Pixel InitiateCheckout

dja:
**What does this PR do?**
It adds the recommended standard content_type of `product` for `InitiateCheckout`, per FB documentation: https://developers.facebook.com/docs/facebook-pixel/reference/

**Are there breaking changes in this PR?**
No.

**Has this been tested end-to-end? Please provide screenshots on how the fix now populates in the end tool. If not, what was done to test?**

**Any background context you want to provide?**

**Is there parity with the server-side/android/iOS integration (if applicable)?**

**Does this require a metadata change? If so, please link the PR from https://github.com/segmentio/destination-scripts.**

**What are the relevant tickets?**

**Link to CC ticket**

**List all the tests accounts you have used to make sure this change works**

**Helpful Docs**

PR #143, opened 2019-06-25
[DEST-236][DEST-682][DEST-777][CC-4860] Facebook Pixel v2.8.0

CarlosMecha:
Bunch of changes all-in-one:
- [DEST-236](https://segment.atlassian.net/browse/DEST-236): Adds custom properties to standard events with a new [setting](https://app.segment.com/partner-portal/integration/facebook-pixel/settings).
- [DEST-682](https://segment.atlassian.net/browse/DEST-682): Allows to override the `content_type` field in the integrations object, as `{ integrations: { 'Facebook Pixel': { contentType: 'myType' } } }`.
- [DEST-777](https://segment.atlassian.net/browse/DEST-777): Adds `content_type` to `Checkout Started` event, as requested in #131 .

The setting and new documentation will remain hidden until this version is released to all customers.  

[CC-4860](https://segment.atlassian.net/browse/CC-4860)

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -132,8 +132,7 @@ FacebookPixel.prototype.track = function(track) {
 
       if (dateFields.indexOf(camel(key)) >= 0) {
         if (is.date(val)) {
-          val = val.toISOString().split('T')[0];
-          acc[key] = val;
+          acc[key] = val.toISOString().split('T')[0];
           return acc;
         }
       }
@@ -255,10 +254,7 @@ FacebookPixel.prototype.productListViewed = function(track) {
     'ViewContent',
     {
       content_ids: contentIds,
-      content_type: this.mappedContentTypesOrDefault(
-        track.category(),
-        contentType
-      )
+      content_type: this.getContentType(track, contentType)
     },
     { eventID: track.proxy('messageId') }
   );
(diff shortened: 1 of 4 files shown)


integrations/facebook-pixel/lib/index.js
```diff
@@ -514,20 +500,42 @@ FacebookPixel.prototype.checkoutStarted = function(track) {
...
+ *
+ * @return Content Type array as defined in:
+ * - https://developers.facebook.com/docs/facebook-pixel/reference/#object-properties
+ * - https://developers.facebook.com/docs/marketing-api/dynamic-ads-for-real-estate/audience
  */
-FacebookPixel.prototype.mappedContentTypesOrDefault = function(category, def) {
-  if (!category) return def;
+FacebookPixel.prototype.getContentType = function(track, defaultValue) {
+  // 1- Integration options takes preference over everything
+  var options = track.options('Facebook Pixel');
```
mericsson: Does this code get transpiled? Any reason we use `var` instead of `const`/`let`?

CarlosMecha: It doesn't. `const` and `let` are not supported in old browsers.

integrations/facebook-pixel/lib/index.js
```diff
@@ -578,9 +604,11 @@ FacebookPixel.prototype.formatTraits = function formatTraits(analytics) {
  *
  * @return Payload to send deriveded from the track properties.
  */
-FacebookPixel.prototype.buildPayload = function(track) {
+FacebookPixel.prototype.buildPayload = function(track, isStandardEvent) {
```
mericsson: add jsdoc for `isStandardEvent`?

CarlosMecha: (Added in the next commit ;) )

integrations/facebook-pixel/lib/index.js
mericsson: Got it yep.

PR #144, opened 2019-06-25
Add language & remove tracking pixel & add version

diegoserranoa:
**What does this PR do?**
- Add language to track()
- Remove tracking pixel
- Add segment version

No

We need to receive the language to show the correct strings in the survey. We want to know which version customers are using to debug better any issues.

N/A

If the language is specified in the traits of the track method, it should show the survey in that language.

N/A


CarlosMecha:
Merged as part of #155

PR #149, opened 2019-07-08
[GRWT-240] - Visual Tagger Integration

gurdasnijor:
**What does this PR do?**
https://segment.atlassian.net/browse/GRWT-240

No.

We previously had the integration deployed via the Cavetto pipeline. Due to the end of that project, we've decided to move to the original integration implementation.

https://segment.com/docs/sources/visual-tagger/


integrations/visual-tagger/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@segment/analytics.js-integration-visual-tagger",
+  "version": "0.1.0",
```
gurdasnijor: This is behind the latest published version https://www.npmjs.com/package/@segment/analytics.js-integration-visual-tagger

williamgrosset: My bad not catching this in the review. 🤚

PR #150, opened 2019-07-09
Add Nielsen DTVR module

brennan:
CC Ticket: [CC-4904](https://segment.atlassian.net/browse/CC-4904)

**What does this PR do?**
- This PR adds a new destination component to the analytics.js monorepo: Nielsen DTVR.

- No, changes are net new.

- Nielsen DTVR is a new destination requested by New Fox.
- Project tracked in [GA-873](https://segment.atlassian.net/browse/GA-873).

- n/a

- n/a

- n/a


integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+ */
+
+NielsenDTVR.prototype.sendID3 = function(event) {
+  var id3Events = this.options.sendId3Events;
+  var i;
+  var id3Prop;
+  var id3Tags;
+
+  if (id3Events.length <= 0) return;
+  if (id3Events.length > 0) {
```
CarlosMecha: You won't need this `if` statement if you already checked for the other possible values right before.

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+    for (i = 0; i < id3Events.length; i += 1) {
```
CarlosMecha: you can declare the `var i` inside of the for loop. No need for do it on the top of the function, even if eslint complains.

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+};
+
+/**
+ * Send ID3 tags to Nielsen
+ *
+ * @api private
+ */
+
+NielsenDTVR.prototype.sendID3 = function(event) {
+  var id3Events = this.options.sendId3Events;
```
CarlosMecha: Would this fail if `this.options` is not defined?

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+  var livestream = event.proxy('properties.livestream');
+  var position = event.proxy('properties.position');
+  var time;
+  if (livestream) {
+    time = Date.now(event.timestamp());
+  } else if (position) {
+    time = position;
+  }
+
+  this.client.ggPM('end', time);
```
CarlosMecha: is this expected to be called as `this.client.ggPM('end', null);`?

integrations/nielsen-dtvr/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@segment/analytics.js-integration-nielsen-dtvr",
+  "description": "The Nielsen DTVR analytics.js integration.",
+  "version": "1.0.0",
```
CarlosMecha: let's tag version `0.0.1` until fox has verified that works for them.

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+  var id3Prop;
+  var id3Tags;
+
+  if (id3Events.length <= 0) return;
+  if (id3Events.length > 0) {
+    for (i = 0; i < id3Events.length; i += 1) {
+      if (id3Events[i] === event.event()) {
+        break;
+      }
+      if (i === id3Events.length - 1) {
```
CarlosMecha: This is a little bit confusing. Could we have a `var found = false;` and update in case of found? Then, outside of the for loop, we check `if (!found) return;`

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+  id3Tags = event.proxy('properties.' + id3Prop);
+  if (id3Tags) {
+    // we'll only send ID3 tags to Nielsen if we detect the customer has either
+    // never sent ID3 tags, or if they're different from the previous ID3 tags
+    // sent during the session
+    if (!this.ID3) {
+      this.ID3 = id3Tags;
+      this.client.ggPM('sendID3', this.ID3);
+    } else if (id3Tags !== this.ID3) {
+      this.client.ggPM('sendID3', this.ID3);
```
CarlosMecha: This would be easier to understand if we do something like:
```javascript
if (!this.ID3 || id3Tags !== this.ID3) {
  this.ID3 = id3Tags;
}
this.client.ggPM('sendID3', this.ID3);
```

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+
+  if (loadTypeVal === 'linear') {
+    loadType = '1';
+  } else if (loadTypeVal === 'dynamic') {
+    loadType = '2';
+  }
+
+  return reject({
+    type: 'content',
+    channelName: event.proxy('properties.channel'),
```
CarlosMecha: Is channel name required?

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Map ad properties
+ *
+ * @api private
+ */
+
+NielsenDTVR.prototype.mapAd = function(event) {
+  return reject({
+    type: event.proxy('properties.type'),
```
CarlosMecha: Is `type` and `asset_id` required?

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,347 @@
...
+  ));
+
+/**
+ * Initialize.
+ *
+ * @api public
+ */
+
+NielsenDTVR.prototype.initialize = function() {
+  var protocol = useHttps() ? 'https' : 'http';
```
CarlosMecha: Let's remove that dependency. Use this instead:
```javascript
var protocol = (location.protocol === 'https:' || location.protocol === 'chrome-extension:') ? 'https' : 'http';
```

integrations/nielsen-dtvr/lib/index.js
brennan: Yes - but the integration itself would fail to initialize before we ever got to this point. :)

integrations/nielsen-dtvr/lib/index.js
```diff
@@ -0,0 +1,382 @@
...
+  var adAssetId;
+  var metadata;
+  // Proactively ensure that we call "end" whenever new content
+  // starts. Here, we'll catch it if a customer forgets to call a Segment
+  // "Completed" event, so we'll end the video for them. `end` is also
+  // appropriate during a video interruption,
+  // e.g. if a user is alternating b/w watching two videos on the same page.
+  if (this.previousEvent && track !== this.previousEvent) {
+    if (this.previousEvent.proxy('properties.livestream') === true) {
+      time = this.previousEvent.timestamp().getUTCDate();
```
ccnixon: @brennan - the call to `getUTCDate` will fail if `.timestamp` does not return a `Date` object. Might make sense to typecheck that return value or wrap this in a helper function of some sort.

integrations/nielsen-dtvr/lib/index.js
ccnixon: `options` is built into the `analytics.js-integration` prototype and will always be at least an empty object.

integrations/nielsen-dtvr/lib/index.js
brennan: Yes, added a helper method so that Segment does not send a `loadMetadata` event if all required properties are not present.

integrations/nielsen-dtvr/lib/index.js
brennan: Yes, added a helper method as above.

PR #153, opened 2019-07-17
Removed nested objects from Braze identify and tracking calls

NikoRoberts:
**What does this PR do?**
Stops the plugin from erroring when analytics.track or identify calls are made with incompatible nested object data for Braze.

No

This is related to Segment tech support ticket 353525

Yes, at least iOS does similar sanitization before attempting to send data to Braze
https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/appboy/lib/index.js#L256

No

Segment's own docs cover several times that nested data is not accepted by Braze in these calls.
https://segment.com/docs/destinations/braze/#identify

### DIFF
--- a/integrations/appboy/lib/index.js
+++ b/integrations/appboy/lib/index.js
@@ -312,7 +312,7 @@ Appboy.prototype.identify = function(identify) {
   each(function(value, key) {
     if (
       typeof value === 'object' &&
-      Object.prototype.toString.call(value) !== '[object Array]'
+      Array.isArray(value)
     ) {
       delete traits[key];
     }
@@ -370,7 +370,7 @@ Appboy.prototype.track = function(track) {
   // Remove nested objects as Braze doesn't support nested objects in tracking calls
   // https://segment.com/docs/destinations/braze/#track
   each(function(value, key) {
-    if (typeof value === 'object') {
+    if (val != null && typeof value === 'object') {
       delete properties[key];
     }
   }, properties);
--- a/integrations/appboy/lib/index.js
+++ b/integrations/appboy/lib/index.js
@@ -307,6 +307,17 @@ Appboy.prototype.identify = function(identify) {
     delete traits[key];
   }, reserved);
 
+  // Remove nested hash objects as Braze only supports nested array objects in identify calls
+  // https://segment.com/docs/destinations/braze/#identify
+  each(function(value, key) {
+    if (
+      typeof value === 'object' &&
+      Object.prototype.toString.call(value) !== '[object Array]'
+    ) {
+      delete traits[key];
+    }
+  }, traits);
+
   each(function(value, key) {
     window.appboy.getUser().setCustomUserAttribute(key, value);
   }, traits);
@@ -355,6 +366,14 @@ Appboy.prototype.track = function(track) {
   each(function(key) {
     delete properties[key];
   }, reserved);
+
+  // Remove nested objects as Braze doesn't support nested objects in tracking calls
+  // https://segment.com/docs/destinations/braze/#track
+  each(function(value, key) {
+    if (typeof value === 'object') {
+      delete properties[key];
+    }
+  }, properties);
 
   window.appboy.changeUser(userId);
   window.appboy.logCustomEvent(eventName, properties);
--- a/integrations/appboy/lib/index.js
+++ b/integrations/appboy/lib/index.js
@@ -370,7 +370,7 @@ Appboy.prototype.track = function(track) {
   // Remove nested objects as Braze doesn't support nested objects in tracking calls
   // https://segment.com/docs/destinations/braze/#track
   each(function(value, key) {
-    if (val != null && typeof value === 'object') {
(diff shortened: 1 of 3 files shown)


integrations/appboy/test/index.test.js
```diff
@@ -407,19 +408,23 @@ describe('Appboy', function() {
         analytics.called(window.appboy.logCustomEvent, 'event');
       });
 
-      it('should send all properties', function() {
+      it('should send all properties except nested arrays and hashes', function() {
```
clgeoio:
```suggestion
      it('should send all properties except nested arrays and objects', function() {
```

integrations/appboy/test/index.test.js
```diff
@@ -335,10 +335,11 @@ describe('Appboy', function() {
         );
       });
 
-      it('should handle custom traits of all types', function() {
+      it('should handle custom traits of acceptable types and excludes nested hashes', function() {
```
clgeoio:
```suggestion
      it('should handle custom traits of acceptable types and excludes nested objects', function() {
```

integrations/appboy/lib/index.js
```diff
@@ -307,6 +307,17 @@ Appboy.prototype.identify = function(identify) {
     delete traits[key];
   }, reserved);
 
+  // Remove nested hash objects as Braze only supports nested array objects in identify calls
+  // https://segment.com/docs/destinations/braze/#identify
+  each(function(value, key) {
+    if (
+      typeof value === 'object' &&
+      Object.prototype.toString.call(value) !== '[object Array]'
```
clgeoio: could you use `Array.isArray()`?

integrations/appboy/lib/index.js
```diff
@@ -356,6 +367,14 @@ Appboy.prototype.track = function(track) {
     delete properties[key];
   }, reserved);
 
+  // Remove nested objects as Braze doesn't support nested objects in tracking calls
+  // https://segment.com/docs/destinations/braze/#track
+  each(function(value, key) {
+    if (typeof value === 'object') {
```
clgeoio: `typeof null === 'object'` is true, maybe this would be better:
```
function isObject(val) {
  return val != null && typeof val === 'object' && Array.isArray(val) === false;
}
```

integrations/appboy/lib/index.js
NikoRoberts: I could, but this also supports older browsers and I don't think there is any polyfilling done for these integrations

integrations/appboy/test/index.test.js
NikoRoberts: arrays are not objects?

NikoRoberts: > typeof([])
> "object"

integrations/appboy/lib/index.js
NikoRoberts: maybe this instead? Arrays are also not allowed in tracking calls.
```suggestion
    if (val != null && typeof value === 'object') {
```

integrations/appboy/lib/index.js
gpsamson: We have several other integrations that use `Array.isArray()`, so it should be safe to do so here.

integrations/appboy/lib/index.js
gpsamson: A +1 for @NikoRoberts's suggestion:
```
if (val != null && typeof value === 'object') {
```

integrations/appboy/lib/index.js
NikoRoberts:
```suggestion
      Array.isArray(value)
```

NikoRoberts:
Thanks @gpsamson let us know how we can help!

gpsamson:
One more thing @NikoRoberts. Can you bump the version to `1.10.1` here please: https://github.com/NikoRoberts/analytics.js-integrations/blob/braze-fixes/integrations/appboy/package.json#L4

I'll start validating the changes on our end, but if you want to test them out yourself let me know and I can reach out to you via email.

Thanks!


GPS

NikoRoberts:
@gpsamson done
Yes we are definitely interested in testing, please email me at `niko.roberts (a) airtasker.com`

integrations/appboy/lib/index.js
```diff
@@ -356,6 +367,14 @@ Appboy.prototype.track = function(track) {
     delete properties[key];
   }, reserved);
 
+  // Remove nested objects as Braze doesn't support nested objects in tracking calls
+  // https://segment.com/docs/destinations/braze/#track
+  each(function(value, key) {
+    if (val != null && typeof value === 'object') {
```
gpsamson: Whoops. This should actually be `value` instead of `val`. @NikoRoberts

NikoRoberts:
```suggestion
    if (value != null && typeof value === 'object') {
```

NikoRoberts: Done

PR #158, opened 2019-07-29
[DEST-294][CC-4895] Support revenueType when trackRevenuePerProduct is set

CarlosMecha:
Adds `revenueType` when tracking revenue per product.

- [DEST-294](https://segment.atlassian.net/browse/DEST-294)
- [CC-4895](https://segment.atlassian.net/browse/CC-4895)

### DIFF
--- a/integrations/amplitude/lib/index.js
+++ b/integrations/amplitude/lib/index.js
@@ -242,7 +242,11 @@ function logEvent(track, dontSetRevenue) {
     each(function(value, key) {
       // add query params to either `user_properties` or `event_properties`
       type = value;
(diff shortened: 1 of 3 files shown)


integrations/amplitude/lib/index.js
```diff
@@ -242,7 +242,11 @@ function logEvent(track, dontSetRevenue) {
     each(function(value, key) {
       // add query params to either `user_properties` or `event_properties`
       type = value;
-      type === 'user_properties' ? (params[key] = query) : (props[key] = query);
+      if (type === 'user_properties') {
```
mericsson: nice cleanup

PR #160, opened 2019-07-31
Collect monetateId as trait from cookie, if available

shaungallagher:
**What does this PR do?**

Monetate sets its own ID to track users during a session.  That ID is set as a cookie, `mt.v`.  This PR collects that ID as a trait.

The "Monetate Engine API" Destination Component can then associate the event with the proper user session.  This allows us to better personalize the user's experience (e.g. by targeting behavior that occurred earlier in the session).

No.

Monetate identifies visitors and tracks sessions via the Monetate ID. This ID is generated on the first request to Monetate (if it is not provided as part of the request) and identifies a unique device/browser. In JavaScript tag implementations, the Monetate tag sets the  `mt.v` cookie, which persists this ID for further calls to Monetate.

The Monetate ID looks like this: `5.1879957417.1544197401555`.

It is important that the Monetate ID is used to identify the visitor in all subsequent requests during the session.

This PR has been created to support the in-development "Monetate Engine API" Destination Component, which will receive `track` and `identify` events.  In the case of `track` events in particular, we need to pass the `monetateId` so we can determine the visitor session to which the event should be attributed.

Not applicable.

No.

Our developer documentation is available only to logged-in Monetate users.  The relevant portion of the following document has been provided in the "background context" section:

https://knowledge.monetate.com/hc/en-us/articles/360020800172-Calling-The-Engine-API


integrations/monetate/lib/index.js
```diff
@@ -46,6 +46,14 @@ Monetate.prototype.initialize = function() {
     };
   }
   window.monetateQ = window.monetateQ || [];
+
+  // Get monetateId from mt.v cookie.
+  var monetateIdCookie = document.cookie.match('(^|;) ?mt.v=([^;]*)(;|$)');
```
omegabytes: I'm reading this regex as "pull anything between two semicolons, newlines included"

The below returned as a match when I validated it:
```
;mt.v=5.1879957417.1544197401555;
; mt.v=5.1879957417.1544197401555 console.log(process.env);
```

Adding in a `\w` requires the trailing semicolon to be adjacent to a char, but still allows for:
```
;mt.v=5.1879957417.154419740155 console.log(process.env) var a=0;
```

We'd want this expression to capture only the token, or `mt.v=<token>;`, correct?

omegabytes:
Closing due to no response.

PR #166, opened 2019-08-01
[DEST-852] Amplitude: Add mapping of non-property fields to event_props

gpsamson:
**What does this PR do?**
Implements a new "Append Fields To Event Props" setting that allows customers to map any Segment event field to Amplitude's `event_props` during a track event.

https://segment.atlassian.net/browse/DEST-852

Yes! The new setting is an object where keys are Segment field paths like `context.page.title` and its value is the `event_prop` it should be set under.

Example:
```
appendFieldsToEventProps: {
    'context.page.title': 'pageTitle'
}
```

https://segment.atlassian.net/browse/DEST-852

### DIFF
--- a/integrations/amplitude/HISTORY.md
+++ b/integrations/amplitude/HISTORY.md
@@ -1,3 +1,7 @@
+# 3.1.0 / 2019-08-09
+
+- Add mapping of non-property fields to event_props.
+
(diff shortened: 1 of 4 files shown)


gpsamson:
Tested successfully in staging. Merging and releasing as `1.10.1`!

Issue #110, opened 2019-03-01
[appboy] Only call changeUser when userID exists 


SegmentDestinationsBot:
_Migrated from [#22](https://github.com/segment-integrations/analytics.js-integration-appboy/issues/22) by @dwillislc_

When an identify, orderCompleted, page or track call is made without an userID, the mapping file is still attempting to call changeUser. This is resulting in a non-fatal error in the console log:
![image](https://user-images.githubusercontent.com/17726756/42385674-714cfe2c-810b-11e8-8c4b-4e802e62f194.png)

I see in Segment's documentation that adding a userID is optional for both the Identify and Track call, so it would be great if we could add some conditional logic: 

```
if (userId) {
  	window.appboy.changeUser(userId);
  }
``` 

Thanks much!

PR #167, opened 2019-08-04
Fixing App Boy Braze calls when values are undefined

NikoRoberts:
**What does this PR do?**
Fix for issue #110 
Currently when injecting Braze SDK with Segment dashboard. There are non-fatal errors being reported because user data isn't in the identify or track calls.
This data is optional, so raising an error when it is not provided doesn't seem to make sense.

No

image


NikoRoberts:
@gpsamson done :)

gpsamson:
Awesome. I'll update things on our end first thing tomorrow morning and shoot you an email. Thanks! @NikoRoberts

PR #171, opened 2019-08-07
Braze: Fix tests

gpsamson:
**What does this PR do?**
Fixes broken tests.

No

### DIFF
--- a/integrations/appboy/test/index.test.js
+++ b/integrations/appboy/test/index.test.js
@@ -335,13 +335,9 @@ describe('Appboy', function() {
         );
       });
 
-      it('should handle custom traits of acceptable types and excludes nested hashes', function() {
+      it('should handle string traits', function() {
         analytics.identify('userId', {
-          song: "Who's That Chick?",
-          artists: ['David Guetta', 'Rihanna'],
-          details: { nested: 'object' },
-          number: 16,
-          date: 'Tue Apr 25 2017 14:22:48 GMT-0700 (PDT)'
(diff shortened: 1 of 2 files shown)


integrations/appboy/test/index.test.js
```diff
@@ -335,37 +335,61 @@ describe('Appboy', function() {
...
+        });
+        analytics.called(window.appboy.changeUser, 'userId');
         analytics.called(
           window.appboy.ab.User.prototype.setCustomUserAttribute,
           'date',
           'Tue Apr 25 2017 14:22:48 GMT-0700 (PDT)'
         );
       });
 
+      it('should exclude nested hash traits', function() {
```
williamgrosset: Should this test pass? Looks like the [previous test](https://github.com/segmentio/analytics.js-integrations/blob/a4c8a799d8b73abb21e3fd657bb7b51686d3bfc6/integrations/appboy/test/index.test.js#L338-L367) excluded this nested object.

gpsamson: This test will pass because there is still a bug floating around in the code that isn't fixed yet. [This line](https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/appboy/lib/index.js#L327) should actually be:
```
typeof value === 'object' && !Array.isArray(value)
```

williamgrosset: Ahh gotcha, thanks for the context!

gpsamson:
Closing in favor of #173 !!

PR #172, opened 2019-08-07
DEST-876 appboy (aka braze) use serverWorkerLocation from settings if available

mericsson:
**What does this PR do?**
Use serverWorkerLocation from settings if available.

No.

One customer has service worker in a different location, so we need to allow the customer to tell Braze about that.

Not applicable I believe.

Yes. It will be an optional integration String setting called `serviceWorkerLocation`.

Ticket: https://segment.atlassian.net/browse/DEST-876
appboy web sdk docs: https://js.appboycdn.com/web-sdk/latest/doc/module-appboy.html


mericsson:
Making change here instead: https://github.com/segmentio/analytics.js-integrations/pull/176

PR #174, opened 2019-08-08
[DEST-376][CC-5116][WEBONISE][SEG-3] logPurchase event call inside track for Braze

pankaj-webonise:
**logPurchase event call inside track**

**There is test case failing**
**should handle custom traits of acceptable types and excludes nested hashes**

We have to fix it.

### DIFF
--- a/integrations/appboy/lib/index.js
+++ b/integrations/appboy/lib/index.js
@@ -33,6 +33,7 @@ var Appboy = (module.exports = integration('Appboy')
   .option('trackNamedPages', false)
   .option('customEndpoint', '')
   .option('version', 1)
+  .option('logPurchaseWhenRevenuePresent', false)
   .tag(
     'v1',
     '<script src="https://js.appboycdn.com/web-sdk/1.6/appboy.min.js">'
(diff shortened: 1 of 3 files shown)


integrations/appboy/lib/index.js
```diff
@@ -322,10 +322,7 @@ Appboy.prototype.identify = function(identify) {
   // Remove nested hash objects as Braze only supports nested array objects in identify calls
   // https://segment.com/docs/destinations/braze/#identify
   each(function(value, key) {
-    if (
-      typeof value === 'object' &&
-      Array.isArray(value)
-    ) {
+    if (typeof value === 'object' && Array.isArray(value)) {
```
vjnathe-webonise: Disable the auto linting, let's not change existing code.

vjnathe-webonise:
@pankaj-webonise resolve the PR conflicts, also add the failing test cases for `properties.revenue` and other required properties not being passed.

integrations/appboy/lib/index.js
```diff
@@ -373,30 +373,40 @@ Appboy.prototype.track = function(track) {
...
-      delete properties[key];
-    }
-  }, properties);
 
   if (userId) {
     window.appboy.changeUser(userId);
   }
-  window.appboy.logCustomEvent(eventName, properties);
+  if (properties.revenue) {
+    window.appboy.logPurchase(
```
vjnathe-webonise: @pankaj-webonise seems this needs to be changed to support product array.
`track.products()` gives multiple products and we need to iterate over it.

PR #177, opened 2019-08-09
[DEST-937] Update Fullstory crossorigin flag and add `fs.log

mericsson:
**What does this PR do?**
1. Enable cross-origin access to errors from fs.js
2. Make FS.log not throw an exception if it is called before FS is ready.

No

Changes included in this version:
1. Enable cross-origin access to errors from fs.js
2. Make FS.log not throw an exception if it is called before FS is ready.

Not applicable

No


mericsson:
The last step of this deployment was complete at 15:29 PST today. It should now be rolled out to all customers. Let us know if you see any issues.

PR #189, opened 2019-08-19
vwo/migrate destination codebase

ccnixon:
**What does this PR do?**
Imports the VWO integration code into this repo.

PR #190, opened 2019-08-20
Migrate visual-website-optimizer

CarlosMecha:
Moves the integration Visual Website Optimizer to the monorepo.

### DIFF
--- a/integrations/visual-website-optimizer/package.json
+++ b/integrations/visual-website-optimizer/package.json
@@ -10,9 +10,10 @@
   ],
   "main": "lib/index.js",
   "scripts": {
-    "test": "make test"
+    "test": "karma start",
+    "test:ci": "karma start karma.conf-ci.js"
   },
-  "author": "Segment \u003cfriends@segment.com\u003e",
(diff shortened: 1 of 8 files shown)


integrations/visual-website-optimizer/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/segmentio/analytics.js-integrations.git"
+  },
+  "dependencies": {
+    "@segment/analytics.js-integration": "^3.1.0",
+    "component-each": "^0.2.6",
+    "next-tick": "^0.2.2"
+  },
+  "devDependencies": {
```
ccnixon: @CarlosMecha - I think it's useful to have integrations share the same core dev dependencies if possible (ie karma, mocha, etc). Also, many of these aren't actually required. Here's the updated dev deps in my earlier PR: https://github.com/segmentio/analytics.js-integrations/pull/189/files#diff-73c610cfbf55667abd8aa7f8c836c158R31-R46. The tests all pass fine with them so there should not be any issues.

PR #194, opened 2019-08-21
[DEST-897] added check to unset params and ref on new session

omegabytes:
**What does this PR do?**
Adds check for unsetParamsReferrerOnNewSession

no

Snippet from [amplitude docs](https://developers.amplitude.com/?javascript#platform-specific-settings):

> unsetParamsReferrerOnNewSession
> If false, the existing referrer and utm_parameter values will be carried through each new session. If set to true, the referrerand utm_parameteruser properties, which include referrer, utm_source, utm_medium, utm_campaign, utm_term, and utm_content, will be set to null upon instantiating a new session. Note: This only works if includeReferrer or includeUtm includeUtm are set to true.

--

no

This PR introduces a new setting `unsetParamsReferrerOnNewSession` which tells amplitude to nullify utm/referrer params at the start of each session.

### DIFF
--- a/integrations/amplitude/lib/index.js
+++ b/integrations/amplitude/lib/index.js
@@ -50,6 +50,7 @@ var Amplitude = (module.exports = integration('Amplitude')
   .option('traitsToSetOnce', [])
   .option('traitsToIncrement', [])
   .option('appendFieldsToEventProps', {})
+  .option('unsetParamsReferrerOnNewSession', false)
   .tag('<script src="' + src + '">'));
 
 /**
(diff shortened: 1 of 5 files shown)


integrations/amplitude/lib/index.js
```diff
@@ -97,7 +97,13 @@ Amplitude.prototype.initialize = function() {
       window.amplitude = amplitude;
       when(loaded, function() {
         window.amplitude.runQueuedFunctions();
-        ready();
+        ready(function() {
+          console.log('calling ready');
```
mericsson: Remove `console.log`

integrations/amplitude/lib/index.js
```diff
@@ -106,7 +112,13 @@ Amplitude.prototype.initialize = function() {
   this.load(function() {
     if (window.amplitude.runQueuedFunctions) {
       window.amplitude.runQueuedFunctions();
-      ready();
+      ready(function() {
+        console.log('calling ready');
```
mericsson: Remove `console.log`

mericsson:
Can you add DEST-?? ticket to PR subject to see what this is fixing?

PR #205, opened 2019-09-04
[DEST-1041] Braze: stop removing keywords

ccnixon:
**What does this PR do?**
Updates the `track` handler in the Braze integration to stop stripping "reserved" keywords from the properties passed to `logCustomEvent`. Braze has told us that these property names **are not** actually prohibited when using their web sdk.

### DIFF
--- a/integrations/appboy/test/index.test.js
+++ b/integrations/appboy/test/index.test.js
@@ -432,21 +432,6 @@ describe('Appboy', function() {
         );
       });
 
-      it('should remove reserved keys from custom event properties', function() {
-        analytics.track('event with properties', {
-          product_id: 'noonz',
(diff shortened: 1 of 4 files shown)


integrations/appboy/test/index.test.js
```diff
@@ -432,21 +432,6 @@ describe('Appboy', function() {
         );
       });
 
-      it('should remove reserved keys from custom event properties', function() {
-        analytics.track('event with properties', {
-          product_id: 'noonz',
-          event_name: 'rihanna',
```
brennan: Rihanna is the best.

ccnixon: @anoonan

PR #213, opened 2019-09-12
DEST-594 Category should be always be EnhancedEcommerce

mericsson:
**What does this PR do?**
Category for track enhanced ecommerce event is now always 'EnhancedEcommerce`.

Yes, if customer previously relied on being able to specify a custom category for enhanced ecommerce events those will now be `EnhancedEcommerce`.

https://segment.atlassian.net/browse/DEST-594

Serverside already always uses `EnhancedEcommerce` as its category setting. https://github.com/segmentio/integrations/blob/cf87c8c01317fe7db6b6e920f57edbd20c88b983/integrations/google-analytics/lib/universal/mapper.js#L1026

Yes, adding a new setting `useEnhancedEcommerceCategory`

### DIFF
--- a/integrations/google-analytics/lib/index.js
+++ b/integrations/google-analytics/lib/index.js
@@ -647,7 +647,7 @@ GA.prototype.pushEnhancedEcommerce = function(track, opts, trackerName) {
   var args = reject([
     self._trackerName + 'send',
     'event',
-    track.category() || 'EnhancedEcommerce',
+    'EnhancedEcommerce',
     track.event() || 'Action not defined',
     track.properties().label,
     extend(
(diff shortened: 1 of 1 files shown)


integrations/google-analytics/lib/index.js
```diff
@@ -647,7 +647,7 @@ GA.prototype.pushEnhancedEcommerce = function(track, opts, trackerName) {
   var args = reject([
     self._trackerName + 'send',
     'event',
-    track.category() || 'EnhancedEcommerce',
```
mericsson: We could put this change behind a setting so it is opt in. That seems safer but this seems correct anyways?

mericsson: @CarlosMecha Curious if you have any concerns about this. I think this is the desired functionality but its possible customers are relying on the existing behavior.

CarlosMecha: That makes sense, let's create a new setting!

PR #214, opened 2019-09-13
DEST-797 fb-pixel Use optional `contentIdPreference` when populating `content_ids`

mericsson:
This would allow customer to optionally set a contentIdPreference which would specify which field to use for contentId.

I think it would be a dropdown menu setting with options:
- `productId`
- `id`
- `sku`

TODO:
- [ ] Unit test `getContentIds`

**What does this PR do?**

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -78,8 +78,22 @@ var defaultPiiProperties = [
  * @param {Object} track
  * @returns {string[]} contentIds
  */
-function getContentIds(track) {
+function getContentId(track) {
   return [track.productId() || track.id() || track.sku() || ''];
+}
+
+function getContentIds(products) {
+  var contentIds = foldl(
+    function(acc, product) {
+      var item = new Track({ properties: product });
+      var key = item.productId() || item.id() || item.sku();
+      if (key) acc.push(key);
+      return acc;
+    },
+    [],
+    products
+  );
+  return contentIds;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


integrations/facebook-pixel/lib/index.js
```diff
@@ -226,6 +267,8 @@ FacebookPixel.prototype.productListViewed = function(track) {
     'ViewContent',
     merge(
       {
+        // Note: `contentIds` here does not go through `getContentIds` since current behavior is
```
mericsson: Not sure if this was an intentional omission but I don't want to change the current behavior either.

integrations/facebook-pixel/lib/index.js
```diff
@@ -267,7 +310,7 @@ FacebookPixel.prototype.productViewed = function(track) {
     'ViewContent',
     merge(
       {
-        content_ids: [track.productId() || track.id() || track.sku() || ''],
+        content_ids: [getContentId(track, this.options.contentIdPreference)],
```
mericsson: Open to other name suggestions here.

mericsson:
Another potential option instead of adding another setting is to add `product_id` to the list of options. Apparently customer likely to be OK with:

`product_id || productId || sku || id`

integrations/facebook-pixel/lib/index.js
ccnixon: Name sounds right to me

ccnixon:
@mericsson I'm pretty sure that `track.productId()` will already return `product_id` _or_ `productId`.

PR #217, opened 2019-09-20
[DEST-1068] Ability to use Advanced Matching on External ID

briemcnally:
**What does this PR do?**
Adds functionality to support Advanced Matching on External ID for Facebook Pixel. Since `external_id` is not part of the Segment spec I will introduce a new setting in Partner Portal marked for Client Side only called `keyForExternalId`. Customers can indicated the key in their traits which they would like to map to FB Pixel `external_id`. 

No breaking changes. This has been tested and validated with unit tests and locally using the Golden Compiler in my browser. 

JIRA Ticket: https://segment.atlassian.net/browse/DEST-1068

Customer Commitment: October 1st

We are not currently building support for this SS or on mobile. 

Yes we will be adding a new setting in PP. See the above comment in the description. 

N/A


integrations/facebook-pixel/lib/index.js
```diff
@@ -582,6 +583,10 @@ FacebookPixel.prototype.formatTraits = function formatTraits(analytics) {
       .toLowerCase();
   var state = address.state && address.state.toLowerCase();
   var postalCode = address.postalCode;
+  var external_id; // eslint-disable-line
+  if (this.options.keyForExternalId.length) {
```
mericsson: Interesting use of `.length`. `keyForExternalId` is guaranteed to be a string? I think so.

To be safer can do:

I think you can actually just do `if (this.options.keyForExternalId) {`

Should be equivalent to what you have.

integrations/facebook-pixel/test/index.test.js
```diff
@@ -2554,6 +2554,45 @@ describe('Facebook Pixel', function() {
...
+        ge: 'm',
+        db: '19910113',
+        external_id: '123456789' // eslint-disable-line
+      };
+      var actual = facebookPixel.formatTraits(analytics);
+
+      analytics.deepEqual(expected, actual);
+    });
+
+    it('should reject an external_id if NOT defined in settings', function() {
```
mericsson: An additional test worthwhile to add is one where you do not specify `facebookPixel.options`.

integrations/facebook-pixel/lib/index.js
briemcnally: @mericsson Yes that was a typo, just pushed a commit to resolve.

integrations/facebook-pixel/test/index.test.js
briemcnally: Just added

PR #218, opened 2019-09-24
Seg 6

vjnathe-webonise:
**What does this PR do?**

Support for user defined position for product in product list.

No.

Initially product index + 1 is mapped as product position, now user can define product position. 

Not Applicable.

No.


CarlosMecha:
@vjnathe-webonise I've opened another PR to run the CI steps #219 . We'll take it from here. Thanks!!

PR #219, opened 2019-09-24
[DEST-869][CC-5311][WEBONISE][SEG-6] Add position to Google products

CarlosMecha:
Allow customers to provide the `position` field for products. Changes from #218


integrations/google-analytics/lib/index.js
```diff
@@ -580,6 +580,25 @@ function formatValue(value) {
   return Math.round(value);
 }
 
+function getProductPosition(item, products) {
+  var position = item.properties().position;
+  if (
+    typeof position !== 'undefined' &&
+    !Number.isNaN(Number(position)) &&
+    Number(position) > -1
```
mericsson: Should this also check `Number.isInteger` ?

integrations/google-analytics/lib/index.js
```diff
@@ -580,6 +580,25 @@ function formatValue(value) {
   return Math.round(value);
 }
 
+function getProductPosition(item, products) {
+  var position = item.properties().position;
```
mericsson: This seems safe. But is there ever a time that `position` is a property that is not intended to be used in this way?

CarlosMecha: Not by our spec. This position is an attribute of a a Product.

integrations/google-analytics/lib/index.js
CarlosMecha: Good point, I'll make that extra check befire the release

CarlosMecha: Sadly `Number.isInteger` is not compatible with early versions of Edge :S I'll leave to google to decide what to do with a floating point number.

PR #220, opened 2019-09-25
SEG-8

vjnathe-webonise:
**What does this PR do?**
Added missing properties for facebook pixel standard events.
 
No

There are some discrepancy in segment specs and the facebook object property, we aligned them. 

No.

No.


CarlosMecha:
Thanks @vjnathe-webonise ! I've opened #221. We'll take it from here :)

PR #221, opened 2019-09-25
[DEST-559][CC-5318][WEBONISE][SEG-8] Add Facebook standard events properties to non-mapped events

CarlosMecha:
Adds Facebook standard events properties to non-mapped Segment events (non ecommerce events).

Work submitted on #220.


integrations/facebook-pixel/lib/index.js
```diff
@@ -198,15 +197,22 @@ FacebookPixel.prototype.productListViewed = function(track) {
...
+  var contents = [];
   var products = track.products();
   var customProperties = this.buildPayload(track, true);
 
   // First, check to see if a products array with productIds has been defined.
   if (Array.isArray(products)) {
     products.forEach(function(product) {
-      var productId = product.productId || product.product_id;
+      var track = new Track({ properties: product });
+      var productId = track.productId() || track.id();
```
mericsson: I think this is a change in behavior. `track.productId()` === product_id || productId.

https://sourcegraph.segment.com/github.com/segmentio/facade@56fad152c8858d198c153ededa38b6c7faed5c0c/-/blob/lib/track.js#L97

Looks like old code was `productId || product_id`

mericsson: Perhaps it is a good change in behavior, but wanted to flag it because it could cause issues. Some customers are passing both `product_id` and `productId` and expecting certain behavior. Here is an example: https://segment.atlassian.net/browse/DEST-797
^ in this case, the change here would be fine but perhaps could cause an issue for someone else.

CarlosMecha: Good catch! I'll update the code before the release.

PR #222, opened 2019-09-27
SEG-4 

vjnathe-webonise:
**What does this PR do?**
Group support for mixpanel integration.

No.

Mixpanel recently releases support for groups,but analytics-js mixpanel integration does not support group yet which is adde part of this PR.

No.

Yes,
**groupIdentifierTraits** option needs to be passed for groupKey information provided from Criteo Customer Success Manager or the Mixpanel Sales Team.


CarlosMecha:
Thanks! We'll take it from here. We'll deploy it as part of #224

PR #223, opened 2019-09-27
SEG-9

vjnathe-webonise:
**What does this PR do?**
Align Criteo event properties as per the latest sdk documentation. 

Missing product ID event will not be tracked.

Criteo events not been tracked for some of the customer due to incorrect properties and event name.

No.

No.


CarlosMecha:
Thanks! I've moved the changes to #225 . We'll take care of it!

PR #224, opened 2019-09-27
[DEST-544][CC-5334][WEBONISE][SEG-4] Add group support to Mixpanel

CarlosMecha:
_This work was originally submitted in #222_

Adds groups support.


integrations/mixpanel/lib/index.js
```diff
@@ -37,6 +37,7 @@ var Mixpanel = (module.exports = integration('Mixpanel')
   .option('trackAllPages', false)
   .option('trackNamedPages', false)
   .option('trackCategorizedPages', false)
+  .option('groupIdentifierTraits', [])
```
brennan: Is `groupIdentifierTraits` a new setting? If so, we should should either run a DCS backfill or ensure we don't call any array methods on it until we check that it's not null/undefined and check the type.

integrations/mixpanel/lib/index.js
```diff
@@ -316,6 +317,32 @@ Mixpanel.prototype.alias = function(alias) {
...
+ *
+ * @param {Group} group
+ */
+
+Mixpanel.prototype.group = function(group) {
+  var groupIdentifierTraits = this.options.groupIdentifierTraits;
+  var groupId = group.groupId();
+  var userId = this.analytics.user().id();
+  var traits = group.properties();
+  if (!groupId || !userId || !groupIdentifierTraits.length) {
```
brennan: Prob should not call `length` until we know whether `groupIdentifierTraits` is not null.

integrations/mixpanel/lib/index.js
```diff
@@ -316,6 +317,32 @@ Mixpanel.prototype.alias = function(alias) {
...
+    return;
+  }
+
+  if (traits && Object.keys(traits).length) {
+    for (var ind = 0; ind < groupIdentifierTraits.length; ind++) {
```
mericsson: Consistency fix for this PR, let's rename `ind` to `i`.

integrations/mixpanel/lib/index.js
CarlosMecha: Is not a new setting.

integrations/mixpanel/lib/index.js
CarlosMecha: Uhm, good point. I think it's defined by default on the `option` function.

integrations/mixpanel/lib/index.js
CarlosMecha: Sadly we can't. In old versions of Javascript the for loop doesn't have its own scope. If we call it the same way, we'll be reusing the variable. It's a bad practice and the linter also blames me :D

mericsson: Ooh got it right.

PR #229, opened 2019-10-09
SEG-15 & 17

vjnathe-webonise:
**What does this PR do?**
- Add support for **versionName** setting, this will allow user to specify version.
- Add support for **trackProductsOnce** setting, this will track all products at once.

No.

Earlier each product is being tracked separately from the list of products.

No.

Yes.
**versionName** : defaults to empty.
**trackProductsOnce**: defaults to false.


CarlosMecha:
Thanks @vjnathe-webonise ! We'll take it from here (#246).

PR #230, opened 2019-10-09
SEG-18 & 19

vjnathe-webonise:
**What does this PR do?**
1) Filter A/B type experiments and track them only if setting enabled.
2) Added option to specify single page application. 

No.

Earlier all events is been tracked even if customer don't want them to be tracked which is not of type A/B events.

No.

Yes.
**isSPA**: default to false.
**trackOnlyABExperiments**: defaults to false.


CarlosMecha:
Thanks @vjnathe-webonise, we'll take it from here :) (#247)

PR #231, opened 2019-10-09
[DEST-973] Support merchandizing events, product merchandizing variables, and product eVars

brennan:
https://segment.atlassian.net/browse/DEST-973

**What does this PR do?**
This PR adds support for
- Merchandizing events
    - This is when we attach a custom Adobe event and a corresponding value to `window.s.events`, e.g. `window.s.events='event23=30'`
- Product merchandizing variables
    - e.g. `window.s.products='Games;Monopoly: 3rd Edition;1;19.00;event5=2.5;event23=10'` (`event23` is a product merch variable)
- Product eVars
    - e.g. `window.s.products='Games;Monopoly: 3rd Edition;1;19.00;event5=2.5;eVar33=1'` (`eVar33` is a product eVar)

Adobe documentation here: https://docs.adobe.com/content/help/en/analytics/implementation/javascript-implementation/variables-analytics-reporting/page-variables.html (see "products" heading)

- No, this update does not change any existing functionality, and only _adds_ support for the three variables mentioned above.

- Levi's request.

- Yes, this PR ensures parity with server-side. We do not support merchandizing events, product merchandizing variables, or product eVars on mobile yet.

Yes. This will require that we expose the existing `merchEvents` setting via CDN. The setting is already in use server side, so should not require a backfill. However, before exposing the setting, I will check DCS to ensure it is safe to do so.

- Adobe documentation: https://docs.adobe.com/content/help/en/analytics/implementation/javascript-implementation/variables-analytics-reporting/page-variables.html (see "products" heading)


integrations/adobe-analytics/lib/index.js
```diff
@@ -414,35 +406,43 @@ AdobeAnalytics.prototype.checkoutStarted = function(track) {
...
- * @param {*} msg
- * @param {*} adobeEvent
+ * @param {*} msg              Segment Page or Track payload.
+ * @param {string} adobeEvent  Adobe standard event.
  */
 
 AdobeAnalytics.prototype.processEvent = function(msg, adobeEvent) {
-  var props = msg.properties();
-
-  var products = msg.products();
```
gpsamson: Just want to call out the removal of `msg.products()`. Facade does some fancy stuff behind the scenes (https://github.com/segmentio/facade/blob/a7cc955f57e14373bb4c502a3ea54a96d8aef9bc/lib/track.js#L382-L386).

In the new changes, it looks like we're calling `properties.products` directly without type checking. This _might_ cause issues if the customer's data has some weird value set for `products`, but I wouldn't be too worried. I just wanted to call it out in case there is a bigger concern.

brennan: Good callout - thanks!

brennan: I think this should be ok since we early return if properties.products is not an array: https://github.com/segmentio/analytics.js-integrations/pull/231/files/f7386d3d50e389cbfa9211c2ef754e89dfe768dd#diff-e7435b76f43719e786ff94a43acc1f6eR901

PR #232, opened 2019-10-11
[CC-5449] [DEST-1242] Customer/Nielsen-requested updates

brennan:
https://segment.atlassian.net/browse/DEST-1242
https://segment.atlassian.net/browse/CC-5449

**What does this PR do?**
- This PR implements a few Nielsen-requested tweaks, in addition to supporting `Video Playback Seek Started`, `Video Playback Buffer Started`, and `Video Playback Buffer Completed`.

- No.

- n/a

- Yes, there will be parity amongst Web, iOS and Android after implementing Nielsen's requested updates. All platforms should be updated in the same week (week of October 14).

- n/a

- n/a


integrations/nielsen-dcr/lib/index.js
```diff
@@ -165,18 +165,19 @@ NielsenDCR.prototype.heartbeat = function(assetId, position, options) {
...
   var properties = 'properties.';
-  var assetIdProp = 'asset_id';
-  if (type) {
-    if (type === 'preroll') properties = 'properties.content.';
-    if (type === 'video') assetIdProp = 'content_asset_id';
-  }
-
+  if (type && type === 'preroll') properties = 'properties.content.';
+
+  var customAssetId = this.options.contentAssetIdPropertyName
```
ccnixon: I think it might be more clear to just use a single conditional here instead of the ternary operator:

```
var customAssetId;
if (this.options.contentAssedItPropertyName) {
    customAssetId = track.proxy(properties + this.options.contentAssetIdPropertyName)
}

File: integrations/nielsen-dcr/lib/index.js [hunk 2]
```diff
@@ -165,18 +165,19 @@ NielsenDCR.prototype.heartbeat = function(assetId, position, options) {
 
 NielsenDCR.prototype.getContentMetadata = function(track, type) {
   var properties = 'properties.';
```
ccnixon: I think the variable name here is a bit confusing as `properties` is usually used to indicate an object containing the event properties. Would it make sense to name this something like `propertiesPath`?

integrations/nielsen-dcr/lib/index.js
```diff
@@ -223,20 +224,25 @@ NielsenDCR.prototype.getContentMetadata = function(track, type) {
...
  * @api private
  */
 
 NielsenDCR.prototype.getAdMetadata = function(track) {
   var type = track.proxy('properties.type');
+
   if (typeof type === 'string') type = type.replace('-', '');
 
+  var customAssetId = track.proxy(
+    'properties.' + this.options.adAssetIdPropertyName
```
ccnixon: What happens if this is just an empty string?

brennan: Good catch - def need to fix this. `track.proxy` would return the full properties object in this case.

integrations/nielsen-dcr/lib/index.js
```diff
@@ -409,93 +428,47 @@ NielsenDCR.prototype.videoPlaybackSeekCompleted = function(track) {
...
   clearInterval(this.heartbeatId);
 
-  var position = track.proxy('properties.position');
+  var timestamp = track.timestamp();
   var livestream = track.proxy('properties.livestream');
-
-  // for livestream just send the current utc timestamp
-  if (livestream) position = +Date.now(track.timestamp());
+  var position = livestream
+    ? +Date.now(timestamp)
```
ccnixon: I don't think `Data.now` accepts an argument. This will always just return the current ms since the epoch regardless of the value of `timestamp`

brennan:
@ccnixon Made some changes based on your `Date.now()` comment - would appreciate a re-review if/when you have time. :D

integrations/nielsen-dcr/lib/index.js
```diff
@@ -115,45 +115,34 @@ NielsenDCR.prototype.page = function(page) {
...
-    if (opts.type === 'content' && opts.livestream) {
-      self.currentPosition = new Date(opts.timestamp);
-      var currentTime = self.currentPosition;
-      // for livestream, properties.position is a negative integer representing offset in seconds from current time
-      currentTime.setSeconds(currentTime.getSeconds() + newPosition);
-      self._client.ggPM('setPlayheadPosition', +Date.now(currentTime)); // UTC
-      // increment timestamp by 1 second
-      currentTime.setSeconds(currentTime.getSeconds() + 1);
-    } else if (newPosition < limit) {
+    if (!livestream) {
```
brennan: @brennan This block is unnecessary.

integrations/nielsen-dcr/lib/index.js
```diff
@@ -164,53 +153,57 @@ NielsenDCR.prototype.heartbeat = function(assetId, position, options) {
  */
 
 NielsenDCR.prototype.getContentMetadata = function(track, type) {
-  var properties = 'properties.';
-  var assetIdProp = 'asset_id';
-  if (type) {
-    if (type === 'preroll') properties = 'properties.content.';
-    if (type === 'video') assetIdProp = 'content_asset_id';
```
brennan: @brennan Should we still be differentiating between video and ad in this method?

integrations/nielsen-dcr/lib/index.js
```diff
@@ -530,7 +476,36 @@ function formatAirdate(airdate) {
...
+  if (!position) return date;
+
+  try {
+    if (typeof position !== 'number') {
+      position = parseInt(position, 10); /* eslint-disable-line */
+    }
+  } catch (e) {
+    // if we can't parse position into an Int for some reason, early return
+    // this will mean that the subsequent Nielsen call fails
+    return;
```
brennan: @brennan return `date` as well.

integrations/nielsen-dcr/lib/index.js
```diff
@@ -530,7 +476,36 @@ function formatAirdate(airdate) {
...
   loadType = loadType === 'dynamic' ? '2' : '1';
   return loadType;
 }
+
+/**
+ * Transforms Segment timestamp into Unix date,
+ * seconds since epoch, in UTC. This method also
+ * handles offsets for livestreams, if applicable.
+ *
+ * @param {*} position
```
brennan: @brennan clarify that `position` represents an offset to a live stream.

integrations/nielsen-dcr/lib/index.js
```diff
@@ -164,53 +147,57 @@ NielsenDCR.prototype.heartbeat = function(assetId, position, options) {
  */
 
 NielsenDCR.prototype.getContentMetadata = function(track, type) {
-  var properties = 'properties.';
-  var assetIdProp = 'asset_id';
-  if (type) {
-    if (type === 'preroll') properties = 'properties.content.';
-    if (type === 'video') assetIdProp = 'content_asset_id';
+  var propertiesPath = 'properties.';
+  if (type && type === 'contentEvent') propertiesPath = 'properties.content.';
```
brennan: @briemcnally Tweaked the logic to device whether we should be checking in `properties.content` for metadata. Turns out we do need this check. `contentEvent` includes a pre-roll ad event and "Video Playback Seek Completed/Buffer Completed" and "Video Playback Resumed".

PR #235, opened 2019-10-11
DEST-906 CC-5451 Prepend advertise custom labels

mericsson:
**What does this PR do?**

Prepend `_fp.event.` to custom labels when advertise is enabled.

No. We are now matching functionality as we had previously documented and Quantcast expects to be sent.

[Our documentation](https://segment.com/docs/destinations/quantcast/#labels) and Quantcast confirmed that custom labels should be prepended with _fp.event.

image

This change is to match that expected functionality.

n/a

No

n/a


mericsson:
@brennan mind giving another 👍 here?

(Should be same as https://github.com/segment-integrations/analytics.js-integration-quantcast/pull/15)

PR #246, opened 2019-10-15
[DEST-245][DEST-671][CC-5480] Amplitude's version name and array of products

CarlosMecha:
_Work submitted in #229_

This change adds a new setting `versionName` that customers can use to initialize the Amplitude SDK. Also we provide another setting to send all products from a `Order Completed` event as once. This new feature was added by Amplitude recently.

Both new features are disabled by default.
- [DEST-245](https://segment.atlassian.net/browse/DEST-245)
- [DEST-671](https://segment.atlassian.net/browse/DEST-671)
- [CC-5480](https://segment.atlassian.net/browse/CC-5480)


integrations/amplitude/lib/index.js
```diff
@@ -464,3 +485,16 @@ function mapRevenueAttributes(track) {
     revenue: track.revenue()
   };
 }
+
+// map product for Order Completed event
+// REF: https://segment.com/docs/destinations/amplitude/#order-completed
+function mapProductAttributes(product) {
```
brennan: Should we map falsy values or remove them?

briemcnally: 👆 Wondering if this should be `return reject({....})`?

CarlosMecha: I would prefer not to use a dependency that has been moved to boneyard. We don't use reject in other parts of this destination, so seems ok to leave it as it is.

PR #249, opened 2019-10-16
[DEST-1157] sync node version with ajs-privatee

ccnixon:
**What does this PR do?**
Enforces a specific node version range when any dependencies are installed. Also updates the circleci image we use to align with what's used in analytics.js-private. This is where the production artifacts are generated so maintaining parity with that is critical.


CarlosMecha:
Closing the PR as we are trying a different approach.

Issue #250, opened 2019-10-17
Heap integration defaults to http in moz-extension contexts.


ntinkler-calendly:
I'm using segment within a web-extension codebase that supports both Chrome and Firefox.

The heap integration for segment does not specify a scheme for the injected script:
https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/heap/lib/index.js#L24

When this script is injected into the background script of a Chrome extension, the protocol is correctly resolved as https.

When this script is injected into the background script of a Firefox extension, the scheme reverts to http.
see: https://imgur.com/a/wumClPn

I'd really prefer not to load mixed content in our extension.  I'd also like my analytics code to not be vulnerable to MITM attacks. 

Right now the workaround is ugly - Stick a mutationObserver on the page, wait for analytics.js to inject the heap script, then fix the scheme myself and re-inject.

I'd like to propose a check in the heap integration for the following two protocols:
"moz-extension:"
"chrome-extension:"
If window.location.protocol is either of those values, I believe the integration should specify the scheme as 'https://' instead of leaving it empty and getting mixed behavior across browsers.

I'm happy to put the PR together myself if needed.

ntinkler-calendly:
Digging deeper, seems more appropriate to fix the issue here:
https://github.com/segmentio/analytics.js-integrations/blob/f20cda994a4d878bbc18ab57617919dc0bc7ccb9/integrations/omniture/build.js#L9048

Since that chunk of code is what's causing the root issue and is already working around a variant of this problem for file:// schemes. 

But the scope of the change is much larger.

PR #251, opened 2019-10-17
Add support for moz-extension pages.

ntinkler-calendly:
Loading this in a Firefox extension results in the wrong scheme being applied to injected script tags - http instead of https.  This adds the same check that already exists for chrome-extension pages, but for Firefox.

**What does this PR do?**
Adds support for "moz-extension" pages.

Unknown, but unlikely.

See initial commit message.  

NA

NA

https://github.com/segmentio/analytics.js-integrations/issues/250

### DIFF
--- a/integrations/omniture/build.js
+++ b/integrations/omniture/build.js
@@ -9040,7 +9040,8 @@
 
           var https =
             document.location.protocol === 'https:' ||
-            document.location.protocol === 'chrome-extension:';
+            document.location.protocol === 'chrome-extension:' ||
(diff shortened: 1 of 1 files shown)


ntinkler-calendly:
I'm going to close this since it doesn't seem to be getting picked up.

I'm aware that the Firefox store will reject extensions that dynamically load scripts (like segment and heap) however I believe this PR would still be useful for extensions that are distributed outside of the Mozilla store, particularly through enterprise controls.

PR #260, opened 2019-10-23
Migrate comscore

CarlosMecha:


mericsson:
lgtm, curious do we have a script for doing these?

CarlosMecha:
Yes we do. It was in this repo but for some reason I can't find it anymore, so I only have my local copy.

PR #267, opened 2019-11-19
[DEST-690][SEG-27] Update Hotjar Snippet Version to v6.

brennan:
**What does this PR do?**
- Updates Hotjar Snippet Version to v6.

- No.

- n/a

- n/a

- No.

- https://segment.atlassian.net/browse/DEST-690
- https://webonise.atlassian.net/browse/SEG-27


integrations/hotjar/HISTORY.md
```diff
@@ -1,3 +1,8 @@
+1.2.0 / 2019-10-18
```
brennan: @CarlosMecha Just copied this PR over from Webonise's fork. Any reason we are bumping to 1.2.0 here rather than 1.1.0?

CarlosMecha: yeah, I think the version 1.1.0 is already out, but not documented.

brennan: Weird, definitely not on npm: https://www.npmjs.com/package/@segment/analytics.js-integration-hotjar

PR #280, opened 2019-11-26
[DEST-1426] Add module to support contractor client-side testing.

brennan:
**What does this PR do?**
- This PR introduces a module that makes it easy to build and test any integration in this monorepo.

- n/a

- n/a

- n/a

- No.

- Largely based on https://github.com/segmentio/ajs-compiler


compiler/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+ - This tool builds all A.js-integrations from the local `../integrations` directory into a local version of A.js. 
+ - The tool retrieves integration settings from `http://cdn.segment.com/v1/projects/${writeKey}/settings`. Because we pull settings from the 
+Segment account associated with the `--writeKey` option, you must enable and set up each integration you'd like to test in your Segment dashboard 
+before building a version of A.js locally using this tool.
+ - Once A.js is built, this tool saves a copy of your A.js build in the `./builds` folder. Build names follow the pattern `analytics-${date}.js`, where 
+`date` is the unix timestamp in seconds at which the file was generated.
+ - Next, the tool launches your default browser, serving a sample `index` file from `./src/index.html`. This file includes the Segment snippet and 
+initializes the your just-generated A.js build so you can begin testing right away.
+
```
CarlosMecha: Could we add a section on how to include a new destination here? It's going to be super-helpful for me when migrating new destinations into the monorepo.

brennan: Excellent idea!!!

PR #288, opened 2019-11-27
More tweaks to A.js tester module.

brennan:
**What does this PR do?**
- This PR makes some small tweaks to improve this module's usability.
- TO DO in another PR: Ensure user is on a specific node version before building a.js.

- n/a

- This PR is based on feedback from user testing.

- n/a

- n/a

- n/a


compiler/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node server.js"
+    "clean-ajs": "rm -r ./builds/analytics*.js"
```
CarlosMecha: You can remove the `-r` flag (to avoid deleting folders) and include `--interactive=once` to ask for confirmation to the user (just in case, to avoid big mistakes.

brennan: Good suggestion! I think `interactive=once` may be Linux-specific? I only see the `i` option on Unix/bash.

PR #290, opened 2019-11-28
feat(identify): use on user login instead of profile push for identify

ankit-arora:
This PR replaces the CleverTap profile push function with on user login function to better replicate Segment's  identify functionality.

No. There are no breaking changes. onUserLogin in CleverTap is same as profile push when identities don't change and for anonymous profiles.

This was reported by our client Bolt. We revisited the code and made the relevant fix.

No. This doesn't require any integration settings change.

More info: https://developer.clevertap.com/docs/concepts-user-profiles#section-maintaining-multiple-user-profiles-on-the-same-device


ankit-arora:
Hi,

Any update on reviewing and merging this PR?

Thanks.

PR #294, opened 2019-12-03
DWEB-3

vjnathe-webonise:
**What does this PR do?**
Added support for page events for fullstory.

No.

Fullstory was not tracking page load events.

No.

Yes, **trackAllPages** - defaults to false , **trackNamedPages** - defaults to true, **trackCategorizedPages** - defaults to true.

NA


CarlosMecha:
Thanks @vjnathe-webonise for submitting this! We'll take it from here.

PR #295, opened 2019-12-03
DWEB-5 Allow support for Dates to be send as playload in event

vjnathe-webonise:
**What does this PR do?**
Added support for sending date as object in event payload for braze.

No.

Braze don't have support for object to be send as event payload except date.

No.

No.

https://www.braze.com/docs/developer_guide/platform_integration_guides/web/analytics/tracking_custom_events/


CarlosMecha:
Thanks @vjnathe-webonise for submitting the PR. We'll take it from here.

PR #296, opened 2019-12-03
DWEB-6

vjnathe-webonise:
**What does this PR do?**
Facebook pixel advance matching for external_id from userId or anonymousId if not present under traits.

No.

Earlier external_id only referenced from traits of identity calls.

No.

No.

https://segment.com/docs/destinations/facebook-pixel/#advanced-matching


vjnathe-webonise:
@CarlosMecha seems like PR closed accidentally, reopening.

CarlosMecha:
Thanks! Taking it from here.

PR #297, opened 2019-12-03
[DWEB-3][DEST-1334]

brennan:
Below copied from: https://github.com/webonise/analytics.js-integrations/pull/19

**What does this PR do?**
Added support for page events for fullstory.

No.

Fullstory was not tracking page load events.

No.

Yes, trackAllPages - defaults to false , trackNamedPages - defaults to true, trackCategorizedPages - defaults to true.
*TO DO*: Add these settings in production partner portal.

NA


integrations/fullstory/test/index.test.js
```diff
@@ -179,5 +186,44 @@ describe('FullStory', function() {
...
+
+      it('should not track categorized pages by default', function() {
+        analytics.page('Category', 'Name');
+        analytics.didNotCall(window.FS.event);
+      });
+
+      it('should track categorized pages if enabled', function() {
+        fullstory.options.trackCategorizedPages = true;
+        analytics.page('Category', 'Name');
+        analytics.called(window.FS.event, 'Viewed Category Page');
```
brennan: @CarlosMecha Note this differs from the expected result in the original Webonise PR, e.g that Webonise test would have failed as written: https://github.com/webonise/analytics.js-integrations/blob/DWEB-3/integrations/fullstory/test/index.test.js#L219

Do you remember if the customer wanted "Viewed Category Page" or "Viewed Category Name Page"?

CarlosMecha: They only mentioned `Viewed <Pagename> Page`... How about we do `Viewed Category <Name> Page`?

brennan: Actually, my bad - `Viewed <Category> <Name> Page` would be handled by `trackNamedPages`, so long as the payload includes both a name and a category. So I think we should leave as is.

PR #298, opened 2019-12-04
[Compiler Tool] Improve error handling when we can't fetch settings.

brennan:
This PR:
- Improves error messages
- Only runs `yarn install --no-lockfile` if the output of `git diff --name-only` is truthy.


compiler/scripts/settings.js
```diff
@@ -19,12 +19,11 @@ async function getSourceSettings(writeKey) {
     request.get({ url: url, gzip: true }, (err, _, body) => {
       if (err) {
         reject(err);
+      } else if (body && body.includes('Invalid path or write key provided.')) {
+        console.error('Please make sure your Segment writeKey was entered correctly.')
```
omegabytes: "More info: https://segment.com/docs/connections/find-writekey"

brennan: Fancy. I like it.

PR #302, opened 2019-12-04
[DEST-1512][CC-5740] Add userId to Pendo

CarlosMecha:
Adds userId to Pendo's request.

- [DEST-1512](https://segment.atlassian.net/browse/DEST-1512)
- [CC-5740](https://segment.atlassian.net/browse/CC-5740)


ccnixon:
Tough to tell where the actual functional change is with the migration/linting updates. Do you want to share a link to the section that was updated?

PR #304, opened 2019-12-05
Add Actions

brennan:
**What does this PR do?**
This PR sets up a github that adds the PR "WEBONISE" to PRs submitted against the `master` branch from a Webonise team member.

Some additional thoughts:
- This action is triggered on _every_ PR against master, but bails unless the PR is submitted by a Webonise team member. We check team membership within the action itself.
- For this to work, we'd also need to enable actions for the repo - not sure if there are security concerns around doing so on a public repo.
- We need to figure out whether running this action against every PR to master and enabling/maintaining actions is worth the value add of being auto-notified of PRs. Right now, Webonise manually pings us and/or links to a PR in Jira. This basically automates away that step.

**What value would this functionality add?**
- Labeling PRs will allow us to webhook a notification to Slack letting us know a PR has been opened. The Github<>Slack integration isn't capable of listening for PRs from specific users or teams yet, but it can listen for labels on PRs.

### DIFF
--- a/.github/actions/add-tag/index.js
+++ b/.github/actions/add-tag/index.js
@@ -5,7 +5,7 @@ const {
     GITHUB_TOKEN
 } = process.env
 
-const PR_LABEL = 'TEST'
+const PR_LABEL = 'WEBONISE'
 const REPO_OWNER = 'segmentio'
 const REPO_NAME = 'analytics.js-integrations'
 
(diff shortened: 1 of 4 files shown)


.github/actions/add-tag/index.js
```diff
@@ -0,0 +1,39 @@
...
+const REPO_OWNER = 'segmentio'
+const REPO_NAME = 'analytics.js-integrations'
+
+const octokit = new github.GitHub(GITHUB_TOKEN)
+
+async function run () {
+  const { number, pull_request } = github.context.payload
+
+  // Check who submitted the PR
+  if (!pull_request.user.login.includes('vjnathe-webonise')) {
```
brennan: Would prob be better to check for a branch or PR name pattern, such as `/SEG*`` rather than a person's name.

gpsamson: Is it possible to check if the user is **not** a part of the 1-class-destinations Github team?

gpsamson: Maybe this? https://developer.github.com/v3/teams/members/#get-team-membership

PR #305, opened 2019-12-05
Add Buildkite Config

briemcnally:
**What does this PR do?**

* This PR adds an initial buildkite configuration.
* More information in the [buildkite setup doc here](https://paper.dropbox.com/doc/Buildkite-User-Guide-AjPLxlOQ8lH5t84ONkwB0fVPAg-rgHkGmr6vm1QIcKDq0hJj).

No.

Segment is migrating from Circle CI Enterprise to using Buildkite. We will maintain 2 CI workflows for a.js integrations. Buildkite will be used as our internal CI workflow. Our external CI flow for contractors will run on CirclCI cloud. 

N/A

N/A

*[ Buildkite User Guide](https://paper.dropbox.com/doc/Buildkite-User-Guide--Ap7I4AvZhT0XXyiJ7yBvH0ZZAg-rgHkGmr6vm1QIcKDq0hJj)
*[ Buildkite migration guide](https://paper.dropbox.com/doc/Buildkite-Migration-Guide--ApsmG57GeLq3RRFQT_N9yB1GAQ-KjGact4XknLOeYoLKzEIg)
* Pipeline can be found here: https://buildkite.com/segment/analytics-dot-js-integrations


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,44 @@
+env:
+  SEGMENT_CONTEXTS: "snyk,npm,aws-credentials,ecr,saucelabs"
```
CarlosMecha: I think we are going to need `npm-publish` (I think that's the name). It's a context with write access on NPM.

briemcnally: @CarlosMecha I talked with tooling and updated the npm context to `npm-publish` so it has read and write access, as well as pass the `NPM_TOKEN` env var to the plugin as well in the publish step.

PR #346, opened 2019-12-10
Fix buildkite publish

briemcnally:
Fixes publish step from https://github.com/segmentio/analytics.js-integrations/pull/305. Adds docker image and npm auth login for publishing.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -34,7 +34,7 @@ steps:
   - wait: ~
 
   - label: ":cloud: Publish"
-    # branches: master
+    branches: master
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -34,12 +34,15 @@ steps:
   - wait: ~
 
   - label: ":cloud: Publish"
-    branches: master
+    # branches: master
     commands:
```
gpsamson:
```suggestion
    commands:
         - npm config set "//registry.npmjs.org/:_authToken" $${NPM_TOKEN}
```

PR #364, opened 2019-12-10
Add CircleCI Cloud Config

briemcnally:
**What does this PR do?**

- Adds Circle CI Config for contractors that will run `yarn install`, `yarn lint`, and `yarn test`. 
- If the contributor makes changes to the `.buildkite` config file it will fail 
- If the contributor makes changes to the `yarn.lock`  file it will fail 

- **Will add the `.circleci/config.yml` to the shell script to fail if contributors make changes to that file. For the purposes of this PR if will make the Cloud CI fail, so would like a review prior to adding that change. It will currently fail if changes are made to `.buildkite` or `yarn.lock`.**

No. 

This Circle CI Cloud Workflow will be used for Contractors to be run on Forked PRs. Internal CI Workflows for PRs will be run on Buildkite. Publishing to NPM will be handled by Buildkite. 

N/A 

No

N/A


.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+            #Check the diff on the commits
+            git diff --name-only $COMMITS | cut -d/ -f1 | sort -u > projects
+            echo -e "Modified directories:\n`cat projects`\n"
+
+            # If modified directories contain yarn, or buildkite directories
+            buildall=0
+            for project in `cat projects`; do
+              if [[ ${project} =~ "yarn" || ${project} =~ ".buildkite" ]]; then
+                buildall=1
+                echo -e "Warning. Dependencies or internal configs have change detected."
```
CarlosMecha: This is AMAZING

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+  lint:
+    docker:
+        - image: circleci/node:8-browsers
+    steps:
+      - attach_workspace: { at: . }
+      - restore_cache:
+          key: deps-{{ checksum "yarn.lock" }}
+      - run:
+          name: Run Linter (not enforcing for now)
+          command: yarn lint
```
CarlosMecha: Would be possible to run checkout, the check step, install, linter and tests in the same job? Attaching and detaching workspaces it's extremely slow with CI and I don't think we get much running all this in parallel.

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  setup:
+    docker:
+      - image: circleci/node:8-browsers
+    steps:
+      - checkout
+      - run:
+          name: Determine which projects have changed and trigger the builds
+          command: |
+            ## Indentify head of master
```
omegabytes: Identify

.circleci/config.yml
briemcnally: @CarlosMecha Just updated so it all runs under one job `build_and_test`.

.circleci/config.yml
```diff
@@ -0,0 +1,55 @@
+version: 2
+jobs:
+  build_and_test:
```
CarlosMecha: Last nitpick, could we call this step `build`? Due to legacy naming, the default job on CircleCI is called `build`, and if you use their CLI for local testing, that's the only job that is able to run.

briemcnally:
Note the `.circleci` config will now be failing since I have added `.circleci` to list of directories for the CI to fail on if it is modified. This is necessary to add since in the future we don't want contractors to edit the `.circleci` config but I needed to update it to migrate to the Circle CI Cloud config. I will use admin privileges to squash and merge this PR once buildkite passes. @CarlosMecha  @omegabytes

PR #374, opened 2019-12-14
DWEB-4 Added safety check on event before doing lowercase

vjnathe-webonise:
**What does this PR do?**
Added safety check on **segmentEvent** before applying toLowerCase.

No.

Unsetting destination configuration for page load conversation resulting in the error in the browser console. **Uncaught TypeError: Cannot read property 'toLowerCase' of undefined**

No.

No.

NA


CarlosMecha:
Thanks! Taking it from here.

PR #377, opened 2019-12-17
DWEB-7

vjnathe-webonise:
**What does this PR do?**
Adds support for DBM cookie matching functionality.

No.

NA

NA

Yes, **getDoubleClickId** defaults to **false** and **googleNetworkId** needs to be passed in options

NA


CarlosMecha:
Taking it from here. Thanks!

PR #379, opened 2019-12-17
[DEST-1563][CC-5818] Fix unsetting custom dimensions for SPA pages for Google Analytics

CarlosMecha:
**Changes**
* Makes this feature optional, with a new setting
* Fixes the logic
* Removes unused `eslint` packages from other integrations

**Links**
- [DEST-1563](https://segment.atlassian.net/browse/DEST-1563)
- [CC-5818](https://segment.atlassian.net/browse/CC-5818)


integrations/google-analytics/lib/index.js
```diff
@@ -229,12 +230,18 @@ GA.prototype.page = function(page) {
...
+  // --> resetCustomDimensions: { "dimension1": null }
+  var resetCustomDimensions = {};
+  for (var i = 0; i < opts.resetCustomDimensionsOnPage.length; i++) {
+    var property = opts.resetCustomDimensionsOnPage[i];
+    if (opts.dimensions[property]) {
+      resetCustomDimensions[opts.dimensions[property]] = null;
+    }
   }
-  window.ga(self._trackerName + 'set', customDimension);
+  window.ga(self._trackerName + 'set', resetCustomDimensions);
```
mericsson: I don't know enough about this integration but looks like you're no longer setting `customDimension`. Is that intentional?

CarlosMecha: Yep, the change is to unset (ie set to `null`) the dimensions specified on the new setting. We do that because by default, Google Analytics keeps the dimensions in memory, and even if you don't send them in a new event, they will get added to the call.

This only happens before a page call, for SPA websites.

PR #386, opened 2019-12-19
Use Tracktor 0.8.1 for Visual Tagger Integration

jlee9595:
Pulls in Tracktor change to ignore length of instrumentation spec array


integrations/visual-tagger/lib/index.js
```diff
@@ -12,8 +12,7 @@ var when = require('do-when');
 var Tracktor = (module.exports = integration('Visual Tagger')
   .global('Tracktor')
   .option('workspaceId', '')
-  .option('sourceId', '')
-  .option('instrumentationSpec', []));
```
jlee9595: If we left this line in then everyone would get an empty instrumentationSpec array as a default as soon as this was merged. We want there to be no default value so it's `null` instead

PR #388, opened 2019-12-23
DWEB-12

vjnathe-webonise:
**What does this PR do?**
Support for ISO 8601 date format for dates below 1-Jan-1970.

No.

Previously date conversion logic converts all the dates into unix timestamp but customer.io does not support dates before 1st Jan 1970 into unix timestamp so we sending them into ISO 8601 format. 

No.

No.

https://customer.io/docs/birthday-and-anniversary-campaigns


CarlosMecha:
Thanks! We'll take it from here.

PR #391, opened 2019-12-31
[DEST-1609] Support disabling ad personalization signals.

brennan:
**What does this PR do?**
This PR supports disabling ad personalization signals per Google's docs: https://support.google.com/analytics/answer/9050852?hl=en

- n/a

- n/a

- n/a

- Yes, we will need to add a new setting, `disableAdPersonalization`. No backfill will be required, as the setting will default to `false`.

- JIRA: https://segment.atlassian.net/browse/DEST-1609
- Google Ads documentation: https://support.google.com/analytics/answer/9050852?hl=en
* https://support.google.com/google-ads/answer/9606827?hl=en

### DIFF
--- a/integrations/google-adwords-new/HISTORY.md
+++ b/integrations/google-adwords-new/HISTORY.md
@@ -1,3 +1,8 @@
+1.2.0 / 2020-02-18
+==================
+
+  * Introduces a new setting `disableAdPersonalization` to disable collection of remarketing data for users who do not wish to view personalized ads. When this settiing is `false`, Segment will set `allow_ad_personalization_signals` to false. 
+
 1.1.1 / 2019-12-09
 ==================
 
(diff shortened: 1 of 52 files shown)


integrations/google-adwords-new/HISTORY.md
```diff
@@ -1,3 +1,8 @@
+1.2.0 / 2020-02-18
+==================
+
+  * Introduces a new setting `disableAdPersonalization` to disable collection of remarketing data for users who do not wish to view personalized ads. When this settiing is `false`, Segment will set `allow_ad_personalization_signals` to false.
```
gpsamson: I think this changelog entry doesn't line up with the behavior. When `disableAdPersonalization` is true, we set `allow_ad_personalization_signals` to false.

briemcnally: @gpsamson good catch, fixed.

PR #396, opened 2020-01-06
DWEB-11

vjnathe-webonise:
**What does this PR do?**
Added support for setting custom props field mapping via option **customExperimentProperties** OR **customCampaignProperties**. Its a mapping of key(segment property) and value(optimizely property).

If **customExperimentProperties** OR **customCampaignProperties** are specified it will override the existing props.

NA

NA

Yes, **customExperimentProperties** OR **customCampaignProperties** is the set of key/value pairs where **key**(segment property) and **value**(optimizely property).

NA

### DIFF
--- a/integrations/optimizely/HISTORY.md
+++ b/integrations/optimizely/HISTORY.md
@@ -1,4 +1,9 @@
 
+3.5.0 / 2019-12-28
+==================
+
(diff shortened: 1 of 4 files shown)


CarlosMecha:
Thanks! We'll take it from here.

PR #398, opened 2020-01-07
[LIB-1546] populate locale in segmentio integration

danieljackins:
**What does this PR do?**

This PR populates "locale" in the context with the language of the web browser.

No

This is for https://segment.atlassian.net/browse/LIB-1546

### DIFF
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -283,8 +283,8 @@ Segment.prototype.normalize = function(message) {
   delete msg.options;
   msg.writeKey = this.options.apiKey;
   ctx.userAgent = navigator.userAgent;
-  const locale = navigator.userLanguage || navigator.language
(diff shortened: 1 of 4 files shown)


integrations/segmentio/lib/index.js
```diff
@@ -283,6 +283,10 @@ Segment.prototype.normalize = function(message) {
   delete msg.options;
   msg.writeKey = this.options.apiKey;
   ctx.userAgent = navigator.userAgent;
+  var locale = navigator.userLanguage || navigator.language;
+  if (!ctx.locale && locale !== undefined) {
```
rpaul80: pretty unrealistic scenario, but if someone sets `ctx.locale` to `0` or `false`, `ctx.locale` will evaluate to `true`

danieljackins: fixed!

PR #403, opened 2020-01-15
[DEST-1482] Bump Sentry to version 5.12.1

brennan:
**What does this PR do?**
This PR does two things:
- Bumps Raven to Sentry v 5.12.1
- Updates settings (deprecates old settings; adds some new ones)

The deprecated settings are:
- maxMessageLength

The new settings are:
- `environment` - String value that adds an "environment" tag to Sentry events
- `debugger` - allows customers to see debug output in their browser console to understand what Sentry is doing under the hood; enabling debug mode means events will not be sent to Sentry, just logged locally

New settings have already been added to Segment's partner portal in production and tested locally. I will unhide these settings once this PR is reviewed, merged, and released.

- No, the PR itself contains no breaking changes. In addition, this PR is designed so that users do not notice any difference in functionality after the migration from Raven to Sentry 5.12.1.

Yes, additional potential customer side effects are described here in paper: https://paper.dropbox.com/doc/Migrate-Sentry-destination-to-latest-Sentry-SDK--AshuJuPqjWDKL7H54i96QwMmAg-DVd8BKi52gvNNpxu9HBMH#:uid=017520775066592364534602&h2=Effects-on-customers

Finally, I will update public-facing documentation after this PR is approved, but before we release the changes to all customers.

-n/a

Yes, see above.

- n/a

### DIFF
--- a/integrations/sentry/lib/index.js
+++ b/integrations/sentry/lib/index.js
@@ -7,6 +7,7 @@
 var integration = require('@segment/analytics.js-integration');
 var is = require('is');
 var foldl = require('@ndhoule/foldl');
+
 /**
  * Expose `Sentry` integration.
  */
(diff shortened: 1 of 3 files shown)


integrations/sentry/lib/index.js
```diff
@@ -7,57 +7,100 @@
...
+      // values from `includePaths` tells the Sentry app which frames in a StackTrace to display in the user's
+      // dashboard. we use the Sentry.Integrations.RewriteFrames plugin to check each frame in a stacktrace and reassign
+      // the value of frame.in_app to true/false depending on whether we find a match: https://docs.sentry.io/platforms/javascript/#rewriteframes
+      if (includePaths.length > 0) {
+        config.integrations.push(
+          new window.Sentry.Integrations.RewriteFrames({
+            iteratee: function(frame) {
+              for (var i = 0; i < includePaths.length; i++) {
+                try {
+                  if (frame.filename.match(new RegExp(includePaths[i]))) {
```
mericsson: Do you know how often `iteratee` function is run?

`new RegExp` has performance implications. I'm thinking you could run `new RegExp` outside of the `iteratee` function just once per `includePath` for better performance.

brennan: Yeah, excellent point - I can def run new `RegExp` outside the iteratee function. Thanks!

PR #408, opened 2020-01-21
[STRATCONN-89][DEST-2224] [Google Tag Manager] Add setting to customize dataLayer name.

brennan:
[Original Jira w/ customer request](https://segment.atlassian.net/browse/STRATCONN-89)
[Jira ticket to track release of this PR](https://segment.atlassian.net/browse/DEST-2224)

https://segment.atlassian.net/browse/STRATCONN-89

**What does this PR do?**
- This PR allows customers to enter a custom dataLayer name in Segment settings. If no custom name is set, we default to `dataLayer`, so this PR shouldn't result in breaking changes for customers currently using GTM.

- No. However, this may be a risky change given that we are changing how the GTM `dataLayer` works for all customers. GTM is a popular destination.

- Request from key account Allergan.

- n/a

- Yes, this will require that we add a new setting, `dataLayer` to Partner Portal.

- Yes, you can name the GTM data layer anything you want per [GTM's documentation here](https://developers.google.com/tag-manager/devguide#renaming).


integrations/google-tag-manager/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@segment/analytics.js-integration-google-tag-manager",
   "description": "The Google Tag Manager analytics.js integration.",
-  "version": "2.5.0",
+  "version": "2.6.0-beta",
```
mericsson: Fun, is this for customer to validate before making it 'official'?

integrations/google-tag-manager/lib/index.js
```diff
@@ -53,7 +61,10 @@ GTM.prototype.initialize = function() {
  */
 
 GTM.prototype.loaded = function() {
-  return !!(window.dataLayer && Array.prototype.push !== window.dataLayer.push);
+  return !!(
+    window[this.options.dataLayer] &&
```
mericsson: Assume `loaded` runs after `initialize`? Otherwise `this.options.dataLayer` might be undefined?

mericsson: Maybe add a comment clarifying that... even if obvious to experienced 'integrations authors'.

integrations/google-tag-manager/test/index.test.js
```diff
@@ -11,7 +11,8 @@ describe('Google Tag Manager', function() {
   var gtm;
   var options = {
     containerId: 'GTM-M8M29T',
-    environment: ''
+    environment: '',
+    dataLayer: ''
```
mericsson: Curious why you have to specify dataLayer as `''` vs leaving it undefined? Is this because the integration code initializes it as `''`?

integrations/google-tag-manager/package.json
brennan: Yes!

integrations/google-tag-manager/test/index.test.js
brennan: Could change this value to `dataLayerName` instead. I'm not sure there's a material difference amongst "null", "undefined", or an empty string in this integration since we're only interested in whether it's a truthy or falsy value in the code itself.

mericsson: Ooh I do like `dataLayerName` instead.. more descriptive of the type of variable.

Yeah I agree no material difference in those falsey values for integrations. Just curious why it was added in that case. No biggie.

integrations/google-tag-manager/lib/index.js
brennan: Hmmm my understanding is the `loaded` method is used exclusively in our test suite.

brennan:
Closing this PR due to lack of comms from success/requesting customer.

PR #417, opened 2020-01-30
Update Parse.ly tracker to config-free tracker

joshuarrrr:
**What does this PR do?**
It updates the loading of the Parse.ly tracker from a legacy tag to our modern integration method described here: https://www.parse.ly/help/integration/basic

This loading method reduces total network requests by avoiding the need for a separate bootstrap loader.

No

This brings the Segment integration up to speed with other 3rd-party Parse.ly integrations. I've removed the test that is no longer needed (because there's no longer any need to create an additional `<meta>` element).

I'm not aware of any currently existing server-side/android/iOS components. As this change only involves the preferred method for the JavaScript tracker, it's not applicable to other integrations. But if there are other Parse.ly integration components, I'd like to know about them so we could ensure they're similarly up to date.

No. The `apikey` setting is used directly in the tag, rather than requiring the creation of a dummy `<meta>` element that only existed to pass that value.

https://www.parse.ly/help/integration/basic
https://www.parse.ly/help/integration/alternative/#load_the_tracker_via_javascript

Testing completed successfully on local.


kbourgoin:
Hi there,

I'd like to resubmit this pull request for review. It is currently blocking the ability of several of our customers to integrate Parsely using Segment. These customers include Webflow, Outsystems, and NAB.

### What does this feature/PR do and why was it raised?

This pull request updates the Segment integration to use a supported version of the Parsely integration.  The current implementation is no longer functional.

### Are there any docs surrounding the change?

Integration instructions for Parsely can be found [here](https://www.parse.ly/help/integration/basic).  Note in particular the tag format. The method in the current Segment implementation is `<script src="//d1z2jf7jlzjs58.cloudfront.net/p.js">` . The newer version includes a customer-specific identifier: `<script id="parsely-cfg" src="//cdn.parsely.com/keys/######/p.js"></script>`. This old method, which uses a generic bootstrap script, is no longer supported.

### What will the potential customer impact be as a result of the change?

The Parse.ly integration will start working.

### What will the potential customer impact be if the change is not implemented?

The Parse.ly integration will not work.

### Is there any current architecture being deprecated (e.g. an endpoint that will no longer be supported)? If yes, what is the set date for the deprecation?

The generic bootstrap script was deprecated in 2020.

pooyaj (approved):

PR #420, opened 2020-02-06
[DEST-1703] Nielsen DTVR Feature Updates/Bump to v1.0.0

brennan:
**What does this PR do?**
- This PR updates mapping logic such that Segment only triggers Nielsen SDK events when the video stream's `load_type` is set to 'linear'. This is desired behavior, as Nielsen DTVR ratings are only meant to measure video streams with 'linear' ad load types, per their reps.
- Ticket: https://segment.atlassian.net/browse/DEST-1703

- Yes, this PR changes the default behavior of the destination, but eliminates clearly erroneous behavior.
- Testing completed successfully in locally via compiler.

- Customer request.

- n/a

- n/a

- n/a


briemcnally:
@brennan Stratconn will plan to release these changes on RT on Monday.

PR #422, opened 2020-02-13
GA: Add checkout options to the checkout step completed event hit

gpsamson:
**What does this PR do?**
- [Google Analytics] Add checkout options to the checkout step completed event hit

https://segment.atlassian.net/browse/STRATCONN-41

### DIFF
--- a/integrations/google-analytics/lib/index.js
+++ b/integrations/google-analytics/lib/index.js
@@ -732,7 +732,7 @@ GA.prototype.orderUpdatedEnhanced = function(track) {
 GA.prototype.checkoutStepViewedEnhanced = function(track) {
   var products = track.products();
   var props = track.properties();
-  var options = extractCheckoutOptions(props);
+  var options = extractCheckoutOptions(track);
   var self = this;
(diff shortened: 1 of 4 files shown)


integrations/google-analytics/lib/index.js
```diff
@@ -774,7 +774,10 @@ GA.prototype.checkoutStepCompletedEnhanced = function(track) {
     option: options
   });
 
-  window.ga(self._trackerName + 'send', 'event', 'Checkout', 'Option');
+  window.ga(self._trackerName + 'send', 'event', 'Checkout', 'Option', {
+    step: props.step || 1,
+    option: options
+  });
```
briemcnally: LGTM! Do we know if there is any negative side effects of sending the step and option twice here? I don't see any based off your tests but was just wondering if you noticed anything in your local testing?

gpsamson: Negative, not that I've noticed in my end-to-end tests. I have a customer testing this first so I'll give them some extra time to look at reports and confirm.

PR #423, opened 2020-02-13
[STRATCONN-77] Add Custom Context Data Variable to Heartbeat

briemcnally:
**What does this PR do?**
JIRA: https://segment.atlassian.net/browse/STRATCONN-77
Add support to send Custom Metadata (Context Data) to Adobe Analytics for Heartbeat events. Per Adobe Analytics documentation we can send addtional custom data in these scenarios: 

1. `trackSessionStart(mediaObj, contextData);` for Video Playback Started and Video Playback Resumed events. 
2. `trackEvent(chapterStart, chapterObj, contextData)` for Video Content Started events.  

We will leverage our existing setting Context Data Variables to map properties to custom content to send to AA. 

Deployment plan is to deploy to Fox test source for validation tomorrow (2/12/20) : https://app.segment.com/foxdcg/sources/web_staging/overview

No. This was tested locally using the Golden Compiler. Validation results were shared with Fox to ensure expected Query Parameter outcomes matched expectations. 

Outbound request: 
```
{
  "anonymousId": "a3219017-b2ae-4f10-a625-6957b022660d",
  "context": {
  ......
 },
  "event": "Video Playback Started",
  "properties": {
    "video_asset_title": "The Llama's First Interview Without The Mask",
    "video_genre": "Reality, Game Show, Music",
    "video_primary_business_unit": "fng",
    "video_series_name": "The Masked Singer",
  },
  "receivedAt": "2020-02-12T18:55:58.100Z",
  "sentAt": "2020-02-12T18:55:57.955Z",
  "timestamp": "2020-02-12T18:55:58.091Z",
  "type": "track",
  "userId": null
}
```
Network Request to Heartbeat Tracking URL:

Screen Shot 2020-02-12 at 2 34 39 PM

Network Request to Server URL:
Screen Shot 2020-02-12 at 2 36 38 PM

N/A 

N/A at this time. 

No. 

Additional Background and Context Doc: https://paper.dropbox.com/doc/Adobe-Analytics-A.js-Heartbeat--AuOKIvh6w8PCLv5MiXNp7AJVAg-FAcr9dX5rqZ9RLTg46Bvk

### DIFF
--- a/integrations/adobe-analytics/lib/index.js
+++ b/integrations/adobe-analytics/lib/index.js
@@ -1229,8 +1229,9 @@ function heartbeatSessionStart(track) {
     props.total_length || 0,
     streamType
   );
-  var contextData = {}; // This might be a custom object for the user.
 
+  // Assign custom context data using the Context Data Variables settings and properties in payload.
+  var contextData = createCustomVideoMetadataContext(track, this.options); // This might be a custom object for the user.
   createStandardVideoMetadata(track, mediaObj);
 
   this.mediaHeartbeats[
@@ -1246,6 +1247,11 @@ function heartbeatVideoStart(track) {
   var props = track.properties();
 
   this.mediaHeartbeats[props.session_id || 'default'].heartbeat.trackPlay();
+  // Assign custom metadata using the Context Data Variables settings and properties in payload.
+  var chapterCustomMetadata = createCustomVideoMetadataContext(
+    track,
+    this.options
+  );
 
   if (!this.mediaHeartbeats[props.session_id || 'default'].chapterInProgress) {
     var chapterObj = videoAnalytics.MediaHeartbeat.createChapterObject(
@@ -1266,7 +1272,7 @@ function heartbeatVideoStart(track) {
     this.mediaHeartbeats[props.session_id || 'default'].heartbeat.trackEvent(
       videoAnalytics.MediaHeartbeat.Event.ChapterStart,
       chapterObj,
-      {}
+      chapterCustomMetadata
     );
     this.mediaHeartbeats[
       props.session_id || 'default'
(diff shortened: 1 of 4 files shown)


integrations/adobe-analytics/lib/index.js
```diff
@@ -1456,6 +1462,19 @@ function createStandardVideoMetadata(track, mediaObj) {
   );
 }
 
+function createCustomVideoMetadataContext(track, options) {
```
gpsamson: Should we be applying the `Context Data Property Prefix` setting here?

briemcnally: Hey @gpsamson this one is a bit confusing. So the reason I left it out of the Heartbeat is because for non-heartbeat events we rely on the Context Data Variables setting mappings to identify which properties in the context object a customer also wants to send as a "Context Data Variable". Inherently we will send all the properties in the property object so they don't need to add mappings for properties. The "Prefix" is used in case there is some kind of nesting within the Context object of the payload.  

There is confusion in terms of how the "Context Data Variables" setting interacts with the Adobe Heartbeat implementation. Per this specific request, Fox is expecting that if they indicate mappings in the "Context Data Variables" that for video events using the heartbeat we will pull that data from the **properties object** (not context) and send it to AA as custom video metadata. We currently don't document at all how customers are supposed to sent custom metadata just that they can, so this is definitely one of those areas for improvement. I left the "prefix" out because we are not looking in the context object though since video metadata is sent in properties.

briemcnally: Let me know if I can clarify anything there.

gpsamson: Got it, makes sense. Thanks for clarifying!

PR #427, opened 2020-02-18
Update FullStory recording snippet

djwhatley-fs:
**What does this PR do?**
This is a simple update to the FS snippet code which initializes FullStory.

No breaking changes.

This snippet code has to be updated from time to time, to keep parity with the latest version of the snippet available within the FullStory app.

N/A

No new settings.


brennan:
Thanks, @djwhatley-fs - this seems like a relatively straightforward change. 

However, while we're updating, wondering if you can help me understand [this line in the Fullstory snippet](https://github.com/segmentio/analytics.js-integrations/pull/427/files#diff-0a369420c4609ee1eabb72728dd8ab42R47). It seems that whenever `window['_fs_namespace']` is set, the snippet logs a warning? I'm not sure if this is intentional, but the result is that whenever Fullstory is loaded by Segment, the message `'FullStory namespace conflict. Please set window["_fs_namespace"].'` is logged. I've confirmed Segment implements the [Fullstory snippet per your documentation here](https://help.fullstory.com/hc/en-us/articles/360020623514-How-do-I-get-FullStory-up-and-running-on-my-site-).

Before we cut a new version with these changes, I wan to ensure we address any outstanding issues. Thanks!

djwhatley-fs:
> Thanks, @djwhatley-fs - this seems like a relatively straightforward change.
> 
> However, while we're updating, wondering if you can help me understand [this line in the Fullstory snippet](https://github.com/segmentio/analytics.js-integrations/pull/427/files#diff-0a369420c4609ee1eabb72728dd8ab42R47). It seems that whenever `window['_fs_namespace']` is set, the snippet logs a warning? I'm not sure if this is intentional, but the result is that whenever Fullstory is loaded by Segment, the message `'FullStory namespace conflict. Please set window["_fs_namespace"].'` is logged. I've confirmed Segment implements the [Fullstory snippet per your documentation here](https://help.fullstory.com/hc/en-us/articles/360020623514-How-do-I-get-FullStory-up-and-running-on-my-site-).
> 
> Before we cut a new version with these changes, I wan to ensure we address any outstanding issues. Thanks!

Hm, well the only reason that should happen is if `window.FS` is already defined. I'm not sure exactly why that would be the case, though I have to suspect it's related to the difference in how the snippet code is loaded here vs. normally. Are you saying this happens invariably when deploying with Segment?

brennan:
My apologies, @djwhatley-fs, the above doesn't seem to be an issue in production. I pushed a commit on top of yours and will release sometime this week. Thanks again for the contribution! https://github.com/segmentio/analytics.js-integrations/pull/442/files

PR #432, opened 2020-02-20
[LIB-1619] Fix Wootric initialization

fathyb:
**What does this PR do?**
Update Wootric to use `@segment/analytics.js-integration@3.3.1` to fix initialization issues.

No.

https://segment.atlassian.net/browse/LIB-1619

No.

No.

N/A


.circleci/config.yml
```diff
@@ -83,7 +83,6 @@ jobs:
               if [[ ${project} =~ "yarn" || ${project} =~ ".buildkite" || ${project} =~ ".circleci" ]]; then
                 buildall=1
                 echo -e "Warning. Dependencies or internal configs have change detected."
-                exit 1
```
fathyb: Hey @briemcnally! Could you confirm if `exit 1` was intended? Does this fix looks good to you?

briemcnally: @fathyb, Carlos had originally wanted me to implement this change because we are using the analytics.js-integrations repo has been migrated to Buildkite to use to run our own CI internally (for Segmenters). The Destinations team has a lot of external contributions from contractors, so in that push, we wanted to be able to keep the circle/config when we migrated to Buildkite to test PRs from external contributors. We would want to keep this here so it would fail if the config is changed by external contractors? 

It looks like the buildkite config is failing because of this: https://github.com/segmentio/analytics.js-integrations/blob/aa10d2c750456e3a57a642b92a861fda4a913387/.buildkite/pipeline.yml#L10. Which points to this: https://github.com/segmentio/analytics.js-integrations/blob/aa10d2c750456e3a57a642b92a861fda4a913387/package.json#L15

fathyb: Oh, that makes a lot of sense. Thanks for confirming!

integrations/wootric/test/index.test.js
```diff
@@ -221,7 +221,9 @@ describe('Wootric', function() {
         analytics.assert(window.wootricSettings.properties.property2 === 'bar');
       });
 
-      it('should set suffix _date to property key and convert date if trait is a date', function() {
```
fathyb: This test didn't work before and is only failing now because [we only test commit changes](https://github.com/segmentio/analytics.js-integrations/blob/0460b7f2ee674525651725ecd7f8029f7949161b/package.json#L14). We can verify this by running:
```bash
git checkout origin/master
git clean -fdx
yarn
yarn workspace @segment/analytics.js-integration-wootric run test
```

PR #439, opened 2020-03-04
Add compiler noRebuild configuration, helpful for testing compiler

mericsson:
**What does this PR do?**

```
~/dev/src/github.com/segmentio/analytics.js-integrations/compiler$ ./compile --help
Usage: compile [options]

Options:
  -w, --writeKey <writeKey>    Segment writeKey to for viewing events in the debugger
  -s, --settings <settings>    (Optional) Relative path to custom integrations settings file
  -c, --cdnDomain <cdnDomain>  Segment CDN domain (default: "http://cdn.segment.com")
  -n, --noRebuild              Skips yarn rebuild
  -p, --port <port>            Set a port to serve the local ajs file from (default: 3000)
  -h, --help                   output usage information
```

The yarn rebuild runs whenever there are non `git add`-ed files. I found that when in development this is often the case. So the `-n` option helped me move faster to just restart compiler at times. There could probably be some more work to just do yarn rebuild for the integrations that changed. That would be nice but not sure how to do that.

I tested this change manually. No need for more testing, just the a.js compiler.

### DIFF
--- a/compiler/scripts/settings.js
+++ b/compiler/scripts/settings.js
@@ -1,8 +1,8 @@
 const ejs = require('ejs');
 const { keys, pick, get } = require('lodash')
-var request = require('request');
-var path = require('path')
-var fs = require('fs')
+const request = require('request');
+const path = require('path')
+const fs = require('fs')
 
 /**
  * Retrieves the settings for the provided Segment Write Key.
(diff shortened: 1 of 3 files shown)


compiler/scripts/settings.js
```diff
@@ -114,7 +114,7 @@ async function context(integrationVersions, coreVersion, writeKey, customSetting
 
   const ctx = {};
   const availableIntegrations = readIntegrationsPackages();
-  ctx.enabled = await getEnabledIntegrations(settings, availableIntegrations);
+  ctx.enabled = getEnabledIntegrations(settings, availableIntegrations);
```
mericsson: `getEnabledIntegrations` doesn't return a promise so no need to `await` it.

PR #440, opened 2020-03-10
[STRATCONN-41] GA: Make checkout options optional + support all cases

gpsamson:
**What does this PR do?**
- Makes checkout options optional when sending a `Checkout Step Completed` event
- Supports all property casing types when sending a `Checkout Step Completed` event

No

https://segment.atlassian.net/browse/STRATCONN-41


briemcnally:
🚢

PR #448, opened 2020-04-01
[FullStory] Add missing definition for window._fs_script

djwhatley-fs:
**What does this PR do?**
Add a missing definition for the variable `window._fs_script` which is necessary to properly handle injection of `fs.js` into iframes. Also adds another minor thing that was missing from the snippet.

No, this is a very simple bug fix.

We discovered a CSP issue some customers were having, and after digging in it became apparent that the root cause is that the FS snippet here is missing a variable definition which we previously considered unnecessary due to how we handle adding the script tag in Segment.

N/A

No


djwhatley-fs:
Hey @brennan, sorry to bug you again so soon, but we discovered an issue with the snippet code I recently submitted and need to get another, very small change in. Hope you are doing well and staying safe with everything going on.

brennan:
Hey @marktisham and @djwhatley-fs, and thanks for the PR. I should have time to review and release early next week! Please let me know if you have any questions in the meantime.

djwhatley-fs:
I don't know if anyone is still monitoring this, so it may not matter at all, but... I wasn't really thinking when I pushed to the fork's master just now, and this should really either be a totally new PR, or if someone on Segment's side is amenable, we could re-work this since it's still open.

djwhatley-fs:
Never mind, we have some other things that need to be added as well for new functionality, so we're just going to open a new PR.

PR #444, opened 2020-03-24
Order Completed Events "fan-out"

edubkendo:
In order to make it easier to segment on individual product properties,
order completed events will iterate over the array of products, sending
one Ordered Product event for each product, as well as an Order Completed event.

**What does this PR do?**
Creates parity for web device connection mode with https://github.com/segmentio/analytics-cloud-integrations/pull/64

**Are there breaking changes in this PR?**
No, this PR should only create new events, not change anything else about how events are currently processed.

**Any background context you want to provide?**
https://github.com/segmentio/analytics-cloud-integrations/pull/61

**Is there parity with the server-side/android/iOS integration components (if applicable)?**
This should create parity with the cloud connection mode integration

**Does this require a new integration setting? If so, please explain how the new setting works**
No.

PR #454, opened 2020-04-08
DEST-1966 vero Order Completed Events "fan-out"

mericsson:
**What does this PR do?**

Merging https://github.com/segmentio/analytics.js-integrations/pull/444

In order to make it easier to segment on individual product properties,
order completed events will iterate over the array of products, sending
one Ordered Product event for each product, as well as an Order Completed event.

What does this PR do?
Creates parity for web device connection mode with segmentio/analytics-cloud-integrations#64

No, this PR should only create new events, not change anything else about how events are currently processed.

segmentio/analytics-cloud-integrations#61

This should create parity with the cloud connection mode integration

Does this require a new integration setting? If so, please explain how the new setting works
No.


integrations/vero/lib/index.js
```diff
@@ -108,6 +108,29 @@ Vero.prototype.track = function(track) {
...
+ *
+ * https://www.getvero.com/api/http/#actions
+ * https://github.com/getvero/vero-api/blob/master/sections/js.md#tracking-events
+ *
+ * @api public
+ * @param {Track} track
+ */
+
+Vero.prototype.orderCompleted = function(track) {
+  var products = track.properties().products;
```
omegabytes, line 122: I think I had an occasional issue with `properties()` being undefined over on the client side FWIW.

brennan: Should be ok - `properties` defaults to `{}` if otherwise undefined: https://github.com/segmentio/facade/blob/master/lib/track.js#L460-L472

Should also be able to use `track.products()` for a shortcut, but then would have to check the length of the returned array: https://github.com/segmentio/facade/blob/master/lib/track.js#L382-L386

mericsson:
fyi @edubkendo release planned for this week, will update if that changes.

PR #457, opened 2020-04-15
[STRATCONN-14] DC Floodlight Pass Top Level UserId on Conversion Event

briemcnally:
JIRA: https://segment.atlassian.net/browse/STRATCONN-14
JIRA: https://segment.atlassian.net/browse/STRATCONN-12
JIRA: https://segment.atlassian.net/browse/STRATCONN-86
CC: https://segment.atlassian.net/browse/CC-6331

**What does this PR do?**
**STRATCONN-14**
Currently users are reporting that when they fire an `identify()` event for a user, on subsequent `track()` calls that top level `userId()` is not being sent to DC Floodlight, only the `userId()` nested in properties within the `track()` call itself is being sent. 

This PR adds functionality specifically for `userId`, where if the segmentProp from settings is `{{userId}}` indicating the customer wants to use the top level `userId()` (which would only be present if the customer fired an identify event on the window), then check `self.analytics.user().id()`, otherwise proceed with existing functionality. 

**NOTE** this change is specific for calls where an `identify()` call was previously made and the `userId()` is stored on the analytics object **AND** the customer wants to map the top level `userId()` field as seen from their Segment debugger (which is the `userId()` on the analytics object). 

**STRATCONN-12**
Adds support for products array Custom Variables. Customer confirmed that DCF would expect the products array to pass through as a comma separated array. For example, say you whitelisted products.$.sku  to the custom Floodlight variable u1​ and passed a call like the one below:
```
analytics.track('Checkout Started', {
sku: '1324abcd',
products: [ { sku: 'sku1' } , { sku: 'sku2' } , { sku: 'sku3' } ]
});
```
DCF would expect the the u1​ parameter to be set as a comma-separated array (example:  `[sku1,sku2,sku3]`​).

**STRATCONN-86**
DCF supports two methods of counting: Items Sold and Transaction based. Currently our DCF Integration only supports Items Sold. Current implementation sets tagParams.qty by # of products in products array, falling back to checking for an actual property called quantity.

For this new change we will implement a new setting useTransactionCounting that will default to false as to not cause any regressions for current customers. When this setting is enabled, and the event is a Sales Tag, we will always default to set to tagParams.qty to 1 per the Transaction counting method.

No, this was you cans see changes validated below: 
Screen Shot 2020-04-16 at 12 01 14 PM

N/A

N/A

No

N/A

### DIFF
--- a/integrations/doubleclick-floodlight/lib/index.js
+++ b/integrations/doubleclick-floodlight/lib/index.js
@@ -80,6 +80,7 @@ Floodlight.prototype.track = function(track) {
       mappedEvents.push(item);
     }
   }
+
   var settings = this.options;
 
   // Must have events mapped and DoubleClick Advertiser ID
@@ -105,7 +106,11 @@ Floodlight.prototype.track = function(track) {
 
         if (Array.isArray(segmentProp)) {
           segmentProp = segmentProp.pop();
-          segmentPropValue = find(track.json(), segmentProp);
+          if (segmentProp === 'userId') {
+            segmentPropValue = self.analytics.user().id();
+          } else {
+            segmentPropValue = find(track.json(), segmentProp);
+          }
         } else {
           segmentPropValue = properties[segmentProp];
         }
--- a/integrations/doubleclick-floodlight/lib/index.js
+++ b/integrations/doubleclick-floodlight/lib/index.js
@@ -20,6 +20,7 @@ var Floodlight = (module.exports = integration('DoubleClick Floodlight')
   .option('getDoubleClickId', false)
   .option('googleNetworkId', '')
   .option('segmentWriteKey', '')
+  .option('useTransactionCounting', false)
   .tag(
     'counter',
     '<iframe src="https://{{ src }}.fls.doubleclick.net/activityi;src={{ src }};type={{ type }};cat={{ cat }};dc_lat=;dc_rdid=;tag_for_child_directed_treatment=;ord={{ ord }}{{ customVariables }}?">'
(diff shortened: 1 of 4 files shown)


integrations/doubleclick-floodlight/lib/index.js
```diff
@@ -105,7 +106,11 @@ Floodlight.prototype.track = function(track) {
 
         if (Array.isArray(segmentProp)) {
           segmentProp = segmentProp.pop();
-          segmentPropValue = find(track.json(), segmentProp);
+          if (segmentProp === 'userId') {
```
gpsamson, line 109: Is there any reason we shouldn't support other top-level values via `track.proxy()`?

briemcnally: They are supported by this: `segmentPropValue = find(track.json(), segmentProp);` which is the existing behavior. It is just that `userId` is stored on the analytics object not the track.

gpsamson: Ahh, okay. Got it.

briemcnally: Updated the description of the change change to make it a little more explicit/clear. 👍

briemcnally:
@gpsamson I added support for nested products variables. Ie. when a customer indicated `products.$.price` to be mapped to `u8`. The feature request was that these be sent as a comma separated array. https://segment.atlassian.net/browse/STRATCONN-12

integrations/doubleclick-floodlight/lib/index.js
```diff
@@ -101,11 +102,27 @@ Floodlight.prototype.track = function(track) {
...
+        var segmentPropValue;
+        if (Array.isArray(segmentProp) && segmentProp[0] === 'products') {
+          segmentProp = segmentProp.pop();
+          var productPropArray = [];
+          each(function(product) {
+            if (product[segmentProp]) {
+              productPropArray.push(product[segmentProp]);
+            }
+          }, track.products());
+          segmentPropValue = productPropArray;
```
gpsamson, line 118: Does this need to be joined?

briemcnally: In Stratconn-12 the customer requested that it be sent as a comma separated array and not a string: "DCF would expect the the u1​ parameter to be set as a comma-separated array (example:  [sku1,sku2,sku3]​)." DCF docs are pretty lacking on this topic but happy to follow up with the CSM on expected behavior before deploying.

gpsamson: Got it. Sorry, I misread your comment above.

PR #458, opened 2020-04-16
Gtag.js Follow-ups

gpsamson:
**What does this PR do?**
- Add campaign (UTM) support to page calls
- Removes `Track All Pages` setting for parity with existing GA integration
- Updates the behavior of `Track Named Pages` and `Track Categorized Pages` to be inline with the existing GA integration

Nah

### DIFF
--- a/integrations/gtag/lib/index.js
+++ b/integrations/gtag/lib/index.js
@@ -745,7 +745,8 @@ function setCustomDimensionsAndMetrics(msg, options) {
  * https://support.google.com/analytics/answer/7475939?hl=en#code
  *
  * @api private
- * @param {Track} track
+ * @param {Facade} msg
+ * @param {Object} opts
  */
 function setContentGroups(msg, opts) {
   if (opts.gaContentGroupings && Object.keys(opts.gaContentGroupings).length) {
--- a/integrations/gtag/lib/index.js
+++ b/integrations/gtag/lib/index.js
@@ -23,7 +23,6 @@ var GTAG = (module.exports = integration('Gtag')
   .option('gaWebAppMeasurementId', '')
   .option('awConversionId', '')
   .option('dcFloodLightId', '')
-  .option('trackAllPages', false)
   .option('trackNamedPages', true)
   .option('trackCategorizedPages', true)
   .option('gaOptions', {
@@ -678,14 +677,14 @@ function trackPageViewEvent(page, options) {
   if (options.includeSearch && props.search) {
     str += props.search;
   }
-  if (options.trackAllPages) {
-    push('event', 'page_view', {
-      page_title: name || category,
-      page_location: props.url,
-      page_path: str,
-      non_interaction: nonInteraction
-    });
-  }
+
+  push('event', 'page_view', {
+    page_title: name || category,
+    page_location: props.url,
+    page_path: str,
+    non_interaction: nonInteraction
+  });
+
   if (name && options.trackNamedPages) {
     push('event', 'page_view', {
       page_title: name,
--- a/integrations/gtag/lib/index.js
+++ b/integrations/gtag/lib/index.js
@@ -19,38 +19,33 @@ var extend = require('extend');
 
 var GTAG = (module.exports = integration('Gtag')
   .global('gtagDataLayer')
-  .option('gaWebMeasurementId', '')
-  .option('gaWebAppMeasurementId', '')
   .option('awConversionId', '')
   .option('dcFloodLightId', '')
   .option('trackNamedPages', true)
   .option('trackCategorizedPages', true)
-  .option('gaOptions', {
-    classic: false,
-    enhancedEcommerce: false,
-    setAllMappedProps: true,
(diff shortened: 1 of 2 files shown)


integrations/gtag/lib/index.js
```diff
@@ -668,41 +667,47 @@ GTAG.prototype.timingCompletedEnhanced = function(track) {
...
-    push('event', 'page_view', {
-      page_title: name || category,
-      page_location: props.url,
-      page_path: str,
-      non_interaction: nonInteraction
-    });
-  }
+
+  push('event', 'page_view', {
+    page_title: name || props.title,
```
gpsamson, line 696: This was changed to stay inline with the existing GA integration: https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/google-analytics/lib/index.js#L207

briemcnally:
@gpsamson those changes LGTM.

integrations/gtag/lib/index.js
```diff
@@ -668,41 +667,47 @@ GTAG.prototype.timingCompletedEnhanced = function(track) {
...
+
+  if (campaign && options.gaWebMeasurementId) {
+    push('config', options.gaWebMeasurementId, {
+      campaign: campaign
+    });
+  }
+
+  if (campaign && options.gaWebAppMeasurementId) {
+    push('config', options.gaWebAppMeasurementId, {
+      campaign: campaign
```
briemcnally, line 685: Any reason we are sending the whole campaign object here and not parsing out the details like in google analytics?

gpsamson: Hm, no exact reason. Good call out. Our spec actually calls the UTM keyword `term` instead of `keyword`. I'll push a commit to fix.

integrations/gtag/lib/index.js
```diff
@@ -19,39 +19,33 @@ var extend = require('extend');
...
+  .option('gaEnhancedEcommerce', false)
+  .option('gaAnonymizeIp', false)
+  .option('gaCookieDomain', 'auto')
+  .option('gaEnhancedLinkAttribution', false)
+  .option('gaOptimizeContainerId', '')
+  .option('gaSampleRate', 100)
+  .option('gaSendUserId', false)
+  .option('gaUseAmpClientId', false)
+  .option('gaSiteSpeedSampleRate', 1)
+  .option('gaSetAllMappedProps', false)
```
gpsamson, line 41: Nested objects aren't supported setting types in Partner Portal so all GA options needed to be moved to the top level.

integrations/gtag/lib/index.js
```diff
@@ -19,39 +19,33 @@ var extend = require('extend');
...
   .tag(
     '<script src="//www.googletagmanager.com/gtag/js?id={{ accountId }}&l=gtagDataLayer">'
   ));
 
 GTAG.on('construct', function(Integration) {
   /* eslint-disable */
-  if (Integration.options.gaOptions.classic) {
-      Integration.page = Integration.pageClassic;
-      Integration.orderCompleted = Integration.orderCompletedClassic;
-    } else if (Integration.options.gaOptions.enhancedEcommerce) {
```
gpsamson, line 54: Per a conversation with Brie and Leah about supporting Classic Mode:

> nah let's not

integrations/gtag/lib/index.js
```diff
@@ -109,58 +102,65 @@ GTAG.prototype.initialize = function() {
...
+    //
+    var customMapSwap = {};
+    for (var field in customMap) {
+      if (customMap.hasOwnProperty(field)) {
+        customMapSwap[customMap[field]] = field;
       }
     }
 
+    gaSetting.custom_map = customMapSwap;
+
```
gpsamson, line 126: See comment about the shape of mappings.

integrations/gtag/lib/index.js
```diff
@@ -725,43 +721,62 @@ function trackEnhancedEvent(eventName, payload) {
...
-          'config',
-          options.gaWebAppMeasurementId,
-          merge(
-            {
-              custom_map: customMap
-            },
-            gaOptions.contentGroupings
-          )
-        );
+    }
```
gpsamson, line 743: The section was changed for two reasons:
1. For parity with the existing GA destination.
2. The property names were being mapped, not the actual values.

integrations/gtag/lib/index.js
```diff
@@ -725,43 +721,62 @@ function trackEnhancedEvent(eventName, payload) {
...
+ * @param {Object} options
+ */
+
+function setCustomDimensionsAndMetrics(props, options) {
+  if (options.gaSetAllMappedProps) {
+    // set custom dimension and metrics if present
+    // REF: https://developers.google.com/analytics/devguides/collection/gtagjs/custom-dims-mets
+    // For content grouping
+    // https://support.google.com/analytics/answer/7475939?hl=en#code
+    var customMap = merge(
```
briemcnally, line 734: Nit but should we remove this reference for content groupings here since it. has its own function below?

PR #456, opened 2020-04-13
[STRATCONN-27] Adobe Analytics Merch Events

briemcnally:
JIRA: https://segment.atlassian.net/browse/STRATCONN-27
CC: TODO

**NOTE:** These changes were merged a few months ago by Brennan but because we received no feedback from Levis from UAT, and the changes were blocking the PR was reverted. Please see original PR here: https://github.com/segmentio/analytics.js-integrations/pull/231

**What does this PR do?**
This PR adds support for:

Merchandizing events
* This is when we attach a custom Adobe event and a corresponding value to window.s.events, e.g. window.s.events='event23=30'

Product merchandizing variables
* e.g. window.s.products='Games;Monopoly: 3rd Edition;1;19.00;event5=2.5;event23=10' (event23 is a product merch variable)

Product eVars
* e.g. window.s.products='Games;Monopoly: 3rd Edition;1;19.00;event5=2.5;eVar33=1' (eVar33 is a product eVar)

Adobe documentation here: https://docs.adobe.com/content/help/en/analytics/implementation/javascript-implementation/variables-analytics-reporting/page-variables.html (see "products" heading)

**Are there breaking changes in this PR?**
* No, this update does not change any existing functionality, and only adds support for the three variables mentioned above.

**Any background context you want to provide?**
* This is a Levi's request and will be concurrently deployed with https://github.com/segmentio/analytics.js-integrations/pull/451. Levis will have 4/13-4/24 for User Acceptance Testing and to provide feedback. 

**Is there parity with the server-side/android/iOS integration components (if applicable)?**
Yes, this PR ensures parity with server-side. Each function added has full parity with SS implementation. Merchandising events are not currently supported on mobile. 

**Does this require a new integration setting? If so, please explain how the new setting works**
No the setting `merchEvents` already exists in metadata. We will need to expose it to the CDN in. order for it to work in A.js. 

**Links to helpful docs and other external resources**
* Adobe documentation: https://docs.adobe.com/content/help/en/analytics/implementation/javascript-implementation/variables-analytics-reporting/page-variables.html (see "products" heading)
* Server side implementation: https://github.com/segmentio/integrations/blob/master/integrations/adobe-analytics/lib/mapper.js

PR #468, opened 2020-05-02
Adobe Analytics: Combine all inflight changes + Implement video playback interrupted

gpsamson:
**What does this PR do?**
- Merges #456 and #465 
- Implements an event handler for `Video Playback Interrupted` which calls the media SDK's `trackPause()` method.
- All changes have been tested in stage and locally via compiler. 
- Changes have been validated by UAT by Fox and Levis. 

N/a

Please see PRs that are being merged into this master PR. 

N/A

No

### DIFF
--- a/integrations/adobe-analytics/lib/index.js
+++ b/integrations/adobe-analytics/lib/index.js
@@ -71,6 +71,7 @@ AdobeAnalytics.global('s')
   .option('props', {})
   .option('hVars', {})
   .option('lVars', {})
+  .option('merchEvents', [])
   .option('contextValues', {})
   .option('customDataPrefix', '')
   .option('reportSuiteId', window.s_account)
@@ -111,11 +112,11 @@ AdobeAnalytics.global('s')
   .sOption('usePlugins', true)
   .tag(
     'default',
-    '<script src="//cdn.segment.com/integrations/omniture/AppMeasurement-2.5.0/appmeasurement.js">'
+    '<script src="//cdn.segment.com/integrations/adobe-analytics/appmeasurement-2.20.0.js">'
   )
   .tag(
     'heartbeat',
-    '<script src="//cdn.segment.com/integrations/omniture/AppMeasurement-2.5.0/appmeasurement-heartbeat.js">'
+    '<script src="//cdn.segment.com/integrations/adobe-analytics/appmeasurement-2.20.0-heartbeat.js">'
   );
 
 /**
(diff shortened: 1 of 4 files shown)


integrations/adobe-analytics/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@segment/analytics.js-integration-adobe-analytics",
   "description": "The Adobe Analytics analytics.js integration.",
-  "version": "1.15.2",
+  "version": "1.16.1-rc.2",
```
briemcnally, line 4: Does this need to be `1.16.1-rc.0`?

briemcnally:
LGTM. Just left one comment about the package version.

integrations/adobe-analytics/package.json
gpsamson: Yes! Good catch

integrations/adobe-analytics/lib/index.js
```diff
@@ -292,19 +293,21 @@ AdobeAnalytics.prototype.track = function(track) {
...
-  // otherwise abort
-  var adobeEvent = aliasEvent(track.event(), this.options.events);
-  if (!adobeEvent) return;
+  // Check if Segment event is mapped in settings; if not, noop
+  var isMapped = this.isMapped(eventName);
+  if (!isMapped) {
+    return;
+  }
 
   this.processEvent(track);
```
briemcnally, line 312:
```suggestion
```suggestion
  // 'cart opened' is not covered in our ecommerce spec and was one-off specced for this integration
if (eventName === 'cart opened' || eventName === 'opened cart') {
  this.processEvent(track, 'scOpen');
} else {
  this.processEvent(track);
};
```

Before we release all these changes GA can we commit this change? This will add support for Cart Opened which was raised by Levis. [I see in our SS code](https://github.com/segmentio/integrations/blob/a626606b319defa9169795651fb7f808887bddd3/integrations/adobe-analytics/lib/mapper.js#L45-L50) we added this as one off support since Cart Opened isn't part of our ecommerce spec. This relates to this ticket that was raised by Levis: https://segment.atlassian.net/browse/STRATCONN-178

briemcnally:
@gpsamson added debug flag set to true per AA docs: https://docs.adobe.com/content/help/en/media-analytics/using/sdk-implement/validation/debugging/sdk-debugging.html. Will cut new releasee candidate and pin Fox.

briemcnally:
Last commit just removed my changes to set debug to true.

integrations/adobe-analytics/lib/index.js
gpsamson: That makes sense to me. Could we commit it tomorrow morning (05/05) and pin Levi's source to this new version? At least we could (a) solve the issue (b) test this mashup of all the changes.

PR #477, opened 2020-05-28
feat (Optimizely): Support passing a specific clientInstance to track

jamesopti:
**What does this PR do?**

Adds an additional parameter to users of this integration, enabling them to specify which instance the JS SDK to use as part of a track call. 

**Example Usage:**

```typescript
// Instantiate an Optimizely client
var clientInstance = optimizely.createInstance({ datafile });
...
analytics.track('my_event', {}, {
  Optimizely: {
    clientInstance,
  }
})
```

No - Without this option, the code still uses `window.optimizelyClientInstance`

Heavy users of our product may instantiate multiple client instances in the same global scope of the browser (window). Today, this integration only supports a single client instance.

With this PR, all callers of `.track` can specify which client instance they want to use for tracking.

N/A

No

### DIFF
--- a/integrations/optimizely/lib/index.js
+++ b/integrations/optimizely/lib/index.js
@@ -118,9 +118,11 @@ Optimizely.prototype.track = function(track) {
 
   push(payload);
 
-  var optimizelyClientInstance = window.optimizelyClientInstance;
+  var optimizelyOptions = track.options('Optimizely');
+  // Prefer the client instance specified via options, falling back to the globally defined one.
+  var optimizelyClientInstance =
+    optimizelyOptions.clientInstance || window.optimizelyClientInstance;
   if (optimizelyClientInstance && optimizelyClientInstance.track) {
-    var optimizelyOptions = track.options('Optimizely');
     var userId =
       optimizelyOptions.userId || track.userId() || this.analytics.user().id();
     var attributes =
--- a/integrations/optimizely/test/index.test.js
+++ b/integrations/optimizely/test/index.test.js
@@ -1125,6 +1125,38 @@ describe('Optimizely', function() {
           window.optimizelyClientInstance.track.restore();
         });
 
+        context(
+          'when the Optimizely.clientInstance option is present',
+          function() {
+            beforeEach(function() {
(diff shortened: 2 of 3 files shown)


Danny-Driscoll:
@jamesopti only thing I might change is adding something to clarify that the option is specific to the Full Stack client instance in the name. We're not introducing it here, but generally the ambiguity between web and full stack functionality is already tough to parse and keep track of. i.e. FSclientInstance potentially?

jamesopti:
> @jamesopti only thing I might change is adding something to clarify that the option is specific to the Full Stack client instance in the name. We're not introducing it here, but generally the ambiguity between web and full stack functionality is already tough to parse and keep track of. i.e. FSclientInstance potentially?

I think thats good feedback. Updated!

integrations/optimizely/lib/index.js
```diff
@@ -121,7 +121,7 @@ Optimizely.prototype.track = function(track) {
   var optimizelyOptions = track.options('Optimizely');
   // Prefer the client instance specified via options, falling back to the globally defined one.
   var optimizelyClientInstance =
-    optimizelyOptions.clientInstance || window.optimizelyClientInstance;
+    optimizelyOptions.FSClientInstance || window.optimizelyClientInstance;
```
circAssimilate, line 124: For customers, would `FullStackClientInstance` be more helpful? I think `FS` makes more sense for me internally, but think the slightly longer, more verbose option couldn't hurt.

circAssimilate:
This looks good form my end. Thanks, @jamesopti!

integrations/optimizely/lib/index.js
```diff
@@ -118,9 +118,12 @@ Optimizely.prototype.track = function(track) {
 
   push(payload);
 
-  var optimizelyClientInstance = window.optimizelyClientInstance;
+  var optimizelyOptions = track.options('Optimizely');
```
nchilada, line 121: Does this always resolve to an object? If it ever resolves to `null` or `undefined`, then `optimizelyOptions.FullStackClientInstance` will throw a `TypeError`.

PR #479, opened 2020-06-04
[WIP] feat(optimizely): Expose Edge API to track Edge experiments

patrickshih-optimizely:
**What does this PR do?**
Wraps all `push` calls with a conditional, which if `window.optimizelyEdge` (the edge API) exists on a page, it'll push to `window.optimizelyEdge` instead of `window.optimizely`.

No, we still track and push to `window.optimizely` in Web.

Allows Performance Edge customers to track Performance Edge experiments now. 

N/A

N/A

### DIFF
--- a/integrations/optimizely/HISTORY.md
+++ b/integrations/optimizely/HISTORY.md
@@ -1,3 +1,8 @@
+3.6.0 / 2020-06-04
+==================
+
+  * Exposed `window.optimizelyEdge` so customers with Performance Edge are able to track Performance Edge related experiments in Segment.
+
 
 3.5.0 / 2019-12-28
 ==================
--- a/integrations/optimizely/HISTORY.md
+++ b/integrations/optimizely/HISTORY.md
@@ -1,7 +1,7 @@
 3.6.0 / 2020-06-04
 ==================
 
-  * Exposed `window.optimizelyEdge` so customers with Performance Edge are able to track Performance Edge related experiments in Segment.
+  * Track conversion events using `window.optimizelyEdge` instead of `window.optimizely` if Optimizely Performance Edge is running on the page.
 
 
 3.5.0 / 2019-12-28
--- a/integrations/optimizely/lib/index.js
+++ b/integrations/optimizely/lib/index.js
@@ -1,642 +1,743 @@
-'use strict';
+// 'use strict';
 
-/**
- * Module dependencies.
- */
+// /**
+//  * Module dependencies.
+//  */
 
-var keys = require('@ndhoule/keys');
-var values = require('@ndhoule/values');
-var foldl = require('@ndhoule/foldl');
-var each = require('@ndhoule/each');
-var integration = require('@segment/analytics.js-integration');
-var tick = require('next-tick');
-var push = require('global-queue')('optimizely', { wrap: false });
-var pushEdge = require('global-queue')('optimizelyEdge', { wrap: false });
+// var keys = require('@ndhoule/keys');
+// var values = require('@ndhoule/values');
+// var foldl = require('@ndhoule/foldl');
+// var each = require('@ndhoule/each');
+// var integration = require('@segment/analytics.js-integration');
+// var tick = require('next-tick');
+// var push = require('global-queue')('optimizely', { wrap: false });
+// var pushEdge = require('global-queue')('optimizelyEdge', { wrap: false });
 
-/**
- * Expose `Optimizely` integration.
- */
+// /**
+//  * Expose `Optimizely` integration.
+//  */
 
-var Optimizely = (module.exports = integration('Optimizely')
-  .option('trackCategorizedPages', true)
-  .option('trackNamedPages', true)
-  .option('variations', false) // send data via `.identify()`
-  .option('listen', true) // send data via `.track()`
-  .option('nonInteraction', false)
-  .option('sendRevenueOnlyForOrderCompleted', true)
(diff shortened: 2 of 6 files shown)


integrations/optimizely/lib/index.js
```diff
@@ -9,8 +9,9 @@ var values = require('@ndhoule/values');
 var foldl = require('@ndhoule/foldl');
 var each = require('@ndhoule/each');
 var integration = require('@segment/analytics.js-integration');
-var push = require('global-queue')('optimizely', { wrap: false });
 var tick = require('next-tick');
+var push = require('global-queue')('optimizely', { wrap: false });
+var pushEdge = require('global-queue')('optimizelyEdge', { wrap: false });
```
patrickshih-optimizely, line 14: Tried to do a fancier variable assignment
```
var push = window.optimizelyEdge ? require('global-queue')('optimizelyEdge', { wrap: false }) 
                   : require('global-queue')('optimizely', { wrap: false });
```
But it wasn't passing the unit test I wrote? Would like to hear ideas for testing to get ^ in.

integrations/optimizely/package.json
```diff
@@ -46,6 +46,7 @@
     "karma-sauce-launcher": "^2.0.2",
     "karma-spec-reporter": "^0.0.32",
     "karma-summary-reporter": "^1.6.0",
+    "lerna": "^3.15.0",
```
nchilada, line 49: Seems to be unnecessary, but I'll let the Segment folks weigh in

integrations/optimizely/lib/index.js
```diff
@@ -47,10 +48,17 @@ var optimizelyContext = {
...
+    pushEdge({
+      type: 'integration',
+      OAuthClientId: '5360906403'
+    });
+  } else {
+    push({
+      type: 'integration',
+      OAuthClientId: '5360906403'
+    });
+  }
```
nchilada, line 61: Interesting, this seems to be documented [here](https://docs.developers.optimizely.com/web/docs/technology-partners#section-registration), but was only ever active in _Classic_ Web (see [here](https://github.com/optimizely/optimizely/blob/0ff9c47d7924fc9e2985220d6ddd5a96294b4d15/src/www/js/client/api.js#L992-L1001) and [here](https://github.com/optimizely/optimizely/blob/784edb96f4bc5fbd149d34796676ba8b3e8f4a69/src/www/js/client/rum.js#L179-L180)) and hasn't been ported to X Web/Edge. Doesn't hurt I guess

integrations/optimizely/HISTORY.md
```diff
@@ -1,3 +1,8 @@
+3.6.0 / 2020-06-04
+==================
+
+  * Exposed `window.optimizelyEdge` so customers with Performance Edge are able to track Performance Edge related experiments in Segment.
```
nchilada, line 4:
```suggestion
  * Track conversion events using `window.optimizelyEdge` instead of `window.optimizely` if Optimizely Performance Edge is running on the page.
```

integrations/optimizely/lib/index.js
nchilada: Hmm, I'm guessing these import statements are executed once on the test page before the mocha suite.. So depending on whether `window.optimizelyEdge` is defined, the `push` var would statically refer to `optimizely` or `optimizelyEdge`, causing tests to fail for `optimizelyEdge` or `optimizely` respectively.

Given that `window.optimizelyEdge` probably _isn't_ defined at the beginning, pushes probably all end up going to `optimizely`, causing the `optimizelyEdge` tests to fail.

integrations/optimizely/package.json
patrickshih-optimizely: TBH, I believe it's a mistake on my end

integrations/optimizely/lib/index.js
patrickshih-optimizely: +1 cool, thanks I should've done some due diligence and also checked the codebase! I'll get rid of pushing this if it's edge

integrations/optimizely/lib/index.js
patrickshih-optimizely: >Given that window.optimizelyEdge probably isn't defined

Yup. I'll keep this for now so we can write test cases for Edge. Makes implementing slightly more clunky, but still readable

integrations/optimizely/test/index.test.js
```diff
@@ -1070,6 +1070,29 @@ describe('Optimizely', function() {
...
+            tags: {}
+          });
+          analytics.didNotCall(window.optimizely.push, {
+            type: 'event',
+            eventName: 'event',
+            tags: {}
+          });
+        });
+      });
+
```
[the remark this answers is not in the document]
nchilada: Okay, it looks like we haven't yet added support for _visual_ tags, but custom event tags (and maybe even global tags) passed via the API are probably supported

integrations/optimizely/lib/index.js
nchilada: No worries! I'm not actually sure what to do about this. We no longer have folks actively developing integrations, but maybe if we get back there one day, it might be useful to be collect this data.

I think we could go either way (continue to track, or not) as long as we do the same thing regardless of what mode we're in (Classic Web, X Web, Edge).

nchilada:
I think this has been superseded by #483

patrickshih-optimizely:
Yup, PR is closed

PR #486, opened 2020-06-26
Support Limited Data Use in Facebook Pixel

mloar:
**What does this PR do?**
This PR will make a ```dataProcessingOptions``` call before ```init``` to opt-in to Limited Data Use. It will do this by default.

No.

Facebook has [announced](https://developers.facebook.com/docs/marketing-apis/data-processing-options) a new Limited Data Use feature that is designed to assist with compliance with the California Consumer Privacy Act (CCPA). Data tagged as Limited Data Use will only be processed by Facebook in ways that allow Facebook to fit within the definition of Service Provider under the CCPA, thus avoiding the possibility that website operators using Facebook Pixel could be considered to be "selling" data to Facebook.

During the month of July Facebook will automatically tag events coming from California as Limited Data Use. But starting in August, clients must opt-in to this behavior by making a ```dataProcessingOptions``` call before ```init```.

Facebook Pixel is only a Device - Web integration.

There are similar options in the [Facebook App Events](https://developers.facebook.com/docs/marketing-apis/data-processing-options#app-events-api) and [Facebook Offline Conversions](https://developers.facebook.com/docs/marketing-apis/data-processing-options#server-side-api-and-offline-conversions-api) APIs, which may be of interest to those using those integrations.

This adds the ```limitedDataUse``` setting, a boolean which defaults to true. If this setting is true, a ```dataProcessingOptions``` call will be made instructing Facebook to apply Limited Data Use if Facebook's geolocation determines the client is in the state of California.

### DIFF
--- a/integrations/facebook-pixel/lib/index.js
+++ b/integrations/facebook-pixel/lib/index.js
@@ -30,6 +30,7 @@ var FacebookPixel = (module.exports = integration('Facebook Pixel')
   .option('standardEventsCustomProperties', [])
   .option('keyForExternalId', '')
   .option('userIdAsExternalId', false)
+  .option('limitedDataUse', true)
   .mapping('standardEvents')
   .mapping('legacyEvents')
   .mapping('contentTypes')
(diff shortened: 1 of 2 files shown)


dja:
@mloar Any updates on this PR?

mloar:
No update yet. The CircleCI build appears to have failed because of some strange git error checking out the code rather than my changes, and I don't have access to retry the job.

My company has a paid account, so I'm going to reach out to support to try to move this along.

gpsamson:
Hey @mloar and @dja. I've opened #493 to resolve the issues with the CI system. A lot of our private CI contexts will only run for folks on the Segment team so external PRs tend to fail. 

I'm going to close this PR in favor of #493 so we can move this along the deployment pipeline!

PR #487, opened 2020-06-26
Heap: add support for secureCookie option

chrismear:
**What does this PR do?**

Adds support for passing Heap's `secureCookie` option to `window.heap.load`.

I don't believe so, but I'm not super-familiar with this repo.

Heap have added an optional `secureCookie` option, in response to the SameSite changes instigated by Chromium in 2020.

Users need to set this to `true` if their product is ever served in an iframe, otherwise incorrect session counting will occur in Heap due the SameSite cookie-handling changes. But we can't just set this to `true` for everyone by default, because _that_ causes incorrect session counting in Heap for users whose product is served over both HTTP and HTTPS. So, this needs to be a user-defined option.

Not applicable — this is a setting related to web cookies set by Heap's client-side JS library.

Yes. A new boolean setting, `secureCookie`, defaulted to `false`. Setting this to `true` instructs Heap to set `SameSite=None;Secure` on its cookie(s), which is necessary for Heap to work correctly when the users' product is hosted in an iframe.

https://help.heap.io/getting-started/installation-guides/web/#iframes
https://help.heap.io/getting-started/installation-guides/web/#securecookie
https://docs.heap.io/docs/heapjs-changelog#490---2020-05-11
https://www.chromium.org/updates/same-site


integrations/heap/lib/index.js
```diff
@@ -61,7 +62,7 @@ Heap.prototype.initialize = function() {
     });
   };
 
-  window.heap.load(this.options.appId);
+  window.heap.load(this.options.appId, this.options);
```
chrismear, line 65: I wasn't sure whether it was best to just pass the `options` dictionary straight through like this (making it easier to support future options), or whether it's standard practice to explicitly list the options, something like:

```javascript
window.heap.load(this.options.appId,
  { secureCookie: this.options.secureCookie }
);
```

PR #493, opened 2020-07-14
Support Limited Data Use in Facebook Pixel

gpsamson:
**NOTE** This is a copy of #486! I had to open this PR so that the private CI contexts would resolve.

**What does this PR do?**
This PR will make a ```dataProcessingOptions``` call before ```init``` to opt-in to Limited Data Use. It will do this by default.

No.

Facebook has [announced](https://developers.facebook.com/docs/marketing-apis/data-processing-options) a new Limited Data Use feature that is designed to assist with compliance with the California Consumer Privacy Act (CCPA). Data tagged as Limited Data Use will only be processed by Facebook in ways that allow Facebook to fit within the definition of Service Provider under the CCPA, thus avoiding the possibility that website operators using Facebook Pixel could be considered to be "selling" data to Facebook.

During the month of July Facebook will automatically tag events coming from California as Limited Data Use. But starting in August, clients must opt-in to this behavior by making a ```dataProcessingOptions``` call before ```init```.

Facebook Pixel is only a Device - Web integration.

There are similar options in the [Facebook App Events](https://developers.facebook.com/docs/marketing-apis/data-processing-options#app-events-api) and [Facebook Offline Conversions](https://developers.facebook.com/docs/marketing-apis/data-processing-options#server-side-api-and-offline-conversions-api) APIs, which may be of interest to those using those integrations.

This adds the ```limitedDataUse``` setting, a boolean which defaults to true. If this setting is true, a ```dataProcessingOptions``` call will be made instructing Facebook to apply Limited Data Use if Facebook's geolocation determines the client is in the state of California.

### DIFF
--- a/integrations/facebook-pixel/test/index.test.js
+++ b/integrations/facebook-pixel/test/index.test.js
@@ -178,6 +178,28 @@ describe('Facebook Pixel', function() {
         analytics.called(window.fbq, 'init', options.pixelId, payload);
       });
     });
+
+    describe('#initialize without LDU', function() {
+      before(function() {
+        options.limitedDataUse = false;
+      });
+
+      after(function() {
+        options.limitedDataUse = true;
+      });
+
+      it('should not call dataProcessingOptions if limitedDataUse is false', function() {
+        analytics.stub(window, 'fbq');
+        analytics.initialize();
+        analytics.didNotCall(
+          window.fbq,
+          'dataProcessingOptions',
+          ['LDU'],
+          0,
+          0
+        );
+      });
+    });
   });
 
   describe('loading', function() {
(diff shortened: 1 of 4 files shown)


integrations/facebook-pixel/test/index.test.js
```diff
@@ -147,6 +148,12 @@ describe('Facebook Pixel', function() {
         );
       });
 
+      it('should call dataProcessingOptions if limitedDataUse is enabled', function() {
```
mericsson, line 151: maybe overkill -- but also add test for when it is disabled?

gpsamson: Good call. I'll add one!

PR #496, opened 2020-07-20
[FB Pixel] Add support for preset data processing options

gpsamson:
**What does this PR do?**
Updates FB Pixel's data processing logic to look for preset options on the window object. If valid parameters are set on `window.fbDataProcessingOptions` we'll pass those to the Pixel SDK. If the parameters set on the window objects are undefined or invalid then we'll fallback to these default options: `['LDU'], 0, 0`.

For more information on valid values see: https://developers.facebook.com/docs/marketing-apis/data-processing-options#reference

Nah

server-side and iOS will be updated to support preset options

## Test Plan
- [x] Unit tests
- [x] Testing completed successfully end-to-end in staging

# Rollback Procedure
- Revert this PR

### DIFF
--- a/integrations/facebook-pixel/HISTORY.md
+++ b/integrations/facebook-pixel/HISTORY.md
@@ -1,3 +1,17 @@
+2.11.1/ 2020-07-22
+==================
+
+  * Add support to override the Data Processing Options by pass them in the load options object.
+  ```
+  analytics.load("<writeKey>", {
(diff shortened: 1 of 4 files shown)


gpsamson:
Hey @brennan - just re-requesting your review as we tweaked one line. Instead of reading from the `window` object we're reading from the settings :D

PR #499, opened 2020-07-30
[SatisMeter] Add support for track and group calls

E-Mud:
**What does this PR do?**
  - Adds anonymousId in all calls
  - Adds support for `track` and `group` calls
  - Adds page related properties to `page` calls
  - Adds `source` parameter on `load` to identify analytics.js as the caller

No

No

- SatisMeter browser widget is already updated to support `source` and `anonymousId`.
- SatisMeter server and browser widget will be updated to support the other new calls and parameters.

No

N/A

### DIFF
--- a/integrations/satismeter/HISTORY.md
+++ b/integrations/satismeter/HISTORY.md
@@ -1,3 +1,12 @@
+2.0.2 / 2020-07-30
+==================
+
+  * Send anonymousId in all calls
+  * Support for `track` calls
+  * Support for `group` calls
+  * Support page related properties
+  * Send `source` parameter to identify analytics.js as the caller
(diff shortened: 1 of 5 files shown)


jsedlacek:
Hi, @mericsson @dominicbarnes @mloar,

Jakub from the @satismeter team here.

We have launched the Event-based targeting for our surveys recently. Without client-side events, it's not very useful for our common customers.

You can find more about it here:
https://www.satismeter.com/blog/event-triggered-surveys/

The change is very safe; it just passes more additional calls from Segment to our javascript API.

What can we do to have this PR merged?

pooyaj (approved):

PR #481, opened 2020-06-12
refactor: Clean up the Optimizely integration

nchilada:
## What does this PR do?

* Don't attempt to interface with Optimizely Classic Web, since Classic is finally dead
* Prepare to support Optimizely Edge, an alternative to Optimizely Web
* Drop all references to customCampaignProperties. It seems to have been documented [here](https://segment.com/docs/connections/destinations/catalog/optimizely-web/#settings) but it couldn't possibly have worked.
* Generally refactor the code and tests.

## Are there breaking changes in this PR?

* Technically yes, but effectively no, since the removed functionality hasn't been relevant for a while.

## Any background context you want to provide?

* Our Classic Web product, the predecessor to X Web, is long gone.
* We recently launched Optimizely Performance Edge, which may be implemented instead of Optimizely Web on any given page. We'll extend analytics.js-integrations to add support for it in a followup pull request.

## Is there parity with the server-side/android/iOS integration components (if applicable)?

N/A

## Does this require a new integration setting? If so, please explain how the new setting works

No. On the contrary, the "Custom Experiment Properties" setting described [here](https://segment.com/docs/connections/destinations/catalog/optimizely-web/#settings) is obsolete now that Optimizely Classic is gone (and was also incorrectly documented or implemented), and the "Custom Campaign Properties" setting can also be un-documented because it too was incorrectly implemented and is now being removed.

**Links to helpful docs and other external resources**

I've linked to some API docs from JSDoc comments!

PR #502, opened 2020-08-13
Support a better debugging mode for unit tests

nchilada:
**What does this PR do?**

* Use a visible Chrome browser instead of headless Chrome when running unit tests with `HEADLESS=false`.
* For the Optimizely integration, facilitate this via a new `yarn test:debug` command.

No. I've confirmed `yarn test` command continues to run in headless mode.

I was having trouble debugging some of the tests in #481. This tweak made debugging much easier because it allowed me to step through my code line-by-line using the Chrome inspector. I'll admit there may be a built-in debugging flow that I'm just not aware of.

N/A

No

N/A

### DIFF
--- a/integrations/optimizely/package.json
+++ b/integrations/optimizely/package.json
@@ -11,6 +11,7 @@
   "main": "lib/index.js",
   "scripts": {
     "test": "karma start",
+    "test:debug": "HEADLESS=false karma start",
     "test:ci": "karma start karma.conf-ci.js"
   },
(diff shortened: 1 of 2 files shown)


nchilada:
I know this is a change to the global karma.conf.js, so I totally understand if you all don't want to accept this.

But FYI @patrickshih-optimizely, this may help you debug the stuff you're working on too. `cd integrations/optimizely && yarn test:debug`

PR #504, opened 2020-08-24
Relay custom customer traits to Friendbuy destination

dondi:
**What does this PR do?**

This PR allows users of the Friendbuy destination to relay customized customer properties to Friendbuy, beyond the standard set that is currently accepted.

No.

Friendbuy allows its merchants to add custom traits to their customers when integrating directly with the Friendbuy customer tracking API as part of a feature called widget targeting: http://help.friendbuy.com/en/articles/1916126-widget-targeting

We would like to allow our merchants to pass these custom parameters through the `identify` event to Friendbuy's Destination integration, so that they can use this feature via Segment. The Friendbuy Segment integration currently only accepts Customer ID, Email Address, First Name, and Last Name.  We'd like to be able to allow merchants to pass in custom parameters through the `identify` event, such as age, VIP, etc.

n/a

No.

General customer tracking API: https://developers.friendbuy.com/#tracking

Widget targeting feature: http://help.friendbuy.com/en/articles/1916126-widget-targeting

### DIFF
--- a/integrations/friendbuy/HISTORY.md
+++ b/integrations/friendbuy/HISTORY.md
@@ -1,3 +1,9 @@
+1.0.8 / 2020-08-24
+==================
+
+  * Relay merchant-specific customer traits to the Friendbuy destination
+  * Fix pre-existing linting errors
+
 1.0.7 / 2018-02-23
 ==================
 
--- a/integrations/friendbuy/lib/index.js
+++ b/integrations/friendbuy/lib/index.js
@@ -13,6 +13,41 @@ var normalize = require('to-no-case');
 var is = require('is');
 
 /**
+ * Find any mapped settings to page `name`.
+ *
+ * Given something like this:
+ *
+ * [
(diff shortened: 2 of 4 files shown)


integrations/friendbuy/HISTORY.md
```diff
@@ -1,3 +1,9 @@
+1.0.8 / 2020-08-24
```
mericsson, line 1: This looks like a minor feature. Can we make this 1.1.0?

dondi: Sure, go ahead 👍

integrations/friendbuy/HISTORY.md
```diff
@@ -1,3 +1,9 @@
+1.1.0 / 2020-12-11
```
dondi, line 1: @mericsson apologies—I belatedly realized that I should be the one to change the version 😅 I have now modified it along with the date. Let me know if any further issues arise. Thanks!

PR #506, opened 2020-09-04
[Heap] Add a blacklist to prevent sending traits

mfgea:
Add a blacklist to prevent sending the configured traits in the addUserProperties call

**What does this PR do?**

Adds a config entry in the Heap plugin. It is a list of fields that should be excluded from calls to the addUserProperties method.

No

Some data compliance policies requires that no PII (Personally identifiable information) is sent to subprocessors. This change would allow that, combined with an id that's also non-identifiable.

n/a

Yes. The setting is called `blacklistedTraits`, and it's an array of strings. The description might be: "List of traits (fields) that should be blacklisted from the identify calls to Heap".

https://help.heap.io/data-privacy/data-privacy-faqs/how-do-i-use-heap-to-comply-with-data-privacy-legislation/
https://en.wikipedia.org/wiki/Personal_data

### DIFF
--- a/integrations/heap/HISTORY.md
+++ b/integrations/heap/HISTORY.md
@@ -1,3 +1,7 @@
+2.1.2 / 2020-09-04
+==================
+
+  * Add a blacklist to prevent sending the configured traits in the addUserProperties call
 
 2.1.1 / 2018-07-06
 ==================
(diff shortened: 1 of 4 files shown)


mfgea:
I think the Build process might be broken, as it failed at git cloning.

PR #507, opened 2020-09-04
Allow fetching Mixpanel library from a different URL

henrinormak:
**What does this PR do?**
Allows configuring a custom URL to load Mixpanel.js from

No.

Allows loading Mixpanel from a proxy-server

N/A

There is a new option/setting - `customLibraryUrl`, which if specified is used as the source for the script tag used to load the library.

N/A


henrinormak:
Any updates? Or am I better off figuring out some alternative?

PR #512, opened 2020-09-15
Specify charset in Segment content-type header

acordiner:
**What does this PR do?**

Explicitly specifies a charset in the content-type header for the Segment integration.

No

The Segment integration is sending a JSON payload with a content-type of `text/plain`. According to the HTTP spec, when no explicit charset is provided for `text/plain` payloads, a default charset value of ISO-8859-1 is assumed. However, the JSON payload is being encoded via `XMLHttpRequest.send()`, which uses UTF-8. Therefore, the charset should be explicitly specified as UTF-8.

Unsure if applicable

No

* XMLHttpRequest spec: "If body is a Document, then set request body to body, serialized, converted, and UTF-8 encoded." (https://xhr.spec.whatwg.org/#the-send()-method)
* HTTP spec: "When no explicit charset parameter is provided by the sender, media subtypes of the "text" type are defined to have a default charset value of "ISO-8859-1" when received via HTTP." (https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html)

### DIFF
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -347,7 +347,7 @@ Segment.prototype.ampId = function(ctx) {
 
 Segment.prototype.enqueue = function(path, message, fn) {
   var url = 'https://' + this.options.apiHost + path;
-  var headers = { 'Content-Type': 'text/plain' };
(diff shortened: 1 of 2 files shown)


acordiner:
Any chance to get this merged?

PR #517, opened 2020-09-23
Proposal: allow passing a callback to Amplitude integration

bradvogel:
Hi there - I'm opening this PR just to start a discussion on whether you'd want to take a change like this.

## The problem
We'd like to log an analytics event immediately before navigating a page. Analytics.js' `trackLink()` is insufficient because it merely waits for a 300ms time. We're finding that 300ms isn't long enough for 20% of our analytics events are dropped. Our underlying analytics integration is Amplitude, and Amplitude supports passing a callback.

## The solution (proposed in this incomplete PR)
When using Amplitude specifically, we'd like to be able to pass a callback. This PR shows a very incomplete solution and only exists to start a conversation about this feature request.

### DIFF
--- a/integrations/amplitude/lib/index.js
+++ b/integrations/amplitude/lib/index.js
@@ -238,7 +238,7 @@ Amplitude.prototype.setTraits = function(traits) {
 
 Amplitude.prototype.track = logEvent;
 
-function logEvent(track, dontSetRevenue) {
+function logEvent(track, dontSetRevenue, callback) {
   this.setDeviceIdFromAnonymousId(track);
 
   var props = track.properties();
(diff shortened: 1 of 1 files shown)


mericsson:
Hey @bradvogel ! :)

Curious if increasing the timeout via `analytics.timeout` would work for you? https://github.com/segmentio/analytics.js-core/blob/d4d70e3ac2e86c1b311b4668d435b41e544281d2/lib/analytics.ts#L738

I haven't tried it but it looks promising? Perhaps the issue then is it affects everywhere else the timeout is used?

mericsson:
Looks like documentation for that is here: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#extending-timeout

bradvogel:
Hey @mericsson ! ;)

Yup, we tried various timeouts ranging from the default of 300 up to 1sec. There was always some % of lost events. Though we'd rather not have any hardcoded, as this flow blocks page navigation and is a poor product experience.

PR #518, opened 2020-09-28
Update PR template with CC note

parlato-segment:
**What does this PR do?**
Adds to the PR template to note the change control requirements

Testing not required because this is a change to the PR template itself. 


gurdasnijor (approved):
🚢

PR #525, opened 2020-12-04
[Hindsight / Rockerbox] Add Custom Domain Option

mericsson:
Merging https://github.com/segmentio/analytics.js-integrations/pull/511 on behalf of @skarim

https://segment.atlassian.net/browse/STRATCONN-509
https://segment.atlassian.net/browse/CC-7107

---

**What does this PR do?**
Adds additional options for users to enter a custom domain & disable cross-domain syncing for their Rockerbox tracker.

**Are there breaking changes in this PR?**
No

**Any background context you want to provide?**
With more browsers blocking third-party cookies, Rockerbox is enabling customers to create a subdomain under their own domain (first-party) that CNAMEs Rockerbox's servers. Rockerbox's Segment integration as currently configured is hardcoded to use a Rockerbox domain and doesn't allow the user to customize this.

Also, please note that "Hindsight" is a legacy product name and the actual integration is now listed as "Rockerbox" in the Segment UI.

**Is there parity with the server-side/android/iOS integration components (if applicable)?**
N/A

**Does this require a new integration setting? If so, please explain how the new setting works**
Yes. The new integration settings would be a text input field named "Custom Tracking Domain" and a switch labeled "Enable Rockerbox Cookie Sync" in the Segment UI.

**Links to helpful docs and other external resources**
Our support doc for Segment is available at https://help.rockerbox.com/docs/segment-pixel-implementation

PR #526, opened 2020-12-07
remove CircleCI Enterprise config

mericsson:
**What does this PR do?**

Removing circleci enterprise config. We have migrated to buildkite and it is no longer necessary. Also it is not working.

More info here 
https://github.com/segmentio/analytics.js-integrations/pull/525#issuecomment-739590064

**Testing**


mericsson:
Didn't realize still used by saas circleci.

PR #540, opened 2020-12-15
Lou/add lou integration

KyleLawson16:
**What does this PR do?**
Adds support for a new integration - [Lou](https://segment.com/docs/connections/destinations/catalog/lou/)

No

**Testing**

We recently built a subscription integration for Lou. This integration will make it even easier for our users to get started by installing the Lou SDK automatically through segment without having to add our Lou script tag manually.

N/A

Yes, I have added a new setting labeled **Organization ID**. This makes the integration available in it's existing form as a basic subscription integration when an `apiKey` setting is included. However, if this new `organizationId` setting is included, it will use this new client-side integration instead.

[Lou Segment Docs](https://www.louassist.com/docs/integrations/segment)


KyleLawson16:
Is there any process that I can go through to get this approved? Our customers are begging us for this integration and they really need it in order to use Segment effectively.

PR #544, opened 2021-01-05
Update FullStory with current snippet and new API usage

djwhatley-fs:
**What does this PR do?**
Updates the FullStory snippet to be in parity with the version currently available in the app. This includes the new `setVars` API, and implements a new default way of tracking pages using that API.

Yes; pages are now tracked via the `setVars` API by default, which may interrupt some users' custom metrics if they are currently depending on custom events for page tracking. However we have added a new option, `trackPagesWithEvents` that re-enables this behavior. If the option is enabled, pages will be sent to FullStory as custom events, in addition to the new method using page variables, which is the recommended option. This is mainly to allow customers who are referencing the custom events some time to transition to the new method.

**Testing**
Testing completed successfully using local Segment test environment. Values were inspected and verified in the browser, and results have been verified in FullStory.

This snippet is long overdue for an update, and we had some customers asking for the `setVars` functionality, so we are doing that in addition to getting the snippet up to date, which is a pretty small change by itself.

N/A

The new setting is described above in the section on breaking changes.

The `FS.setVars` API is not yet publicly documented, but it is available to everyone. The docs will be made available publicly once it is all formally released, but until then you can reference [this doc](https://docs.google.com/document/d/1WfEUoF8kuCPyG0AlWKvCb98NOO8JRnsStKSDbQ6pzQE).


danielstjules:
For this to be a non-breaking change for customers, could we consider making trackPagesWithEvents default to true?

djwhatley-fs:
> For this to be a non-breaking change for customers, could we consider making trackPagesWithEvents default to true?

Sure. After discussing with the team we've agreed that this makes sense.

pooyaj (approved):

PR #547, opened 2021-02-01
Backport analytics-next configuration

nettofarah:
**What does this PR do?**
This change does a couple of things:
- Add analytics-next configuration which allows for releasing individual bundles for destinations in s3
- Add a few missing integrations that are only available at compile time in analytics.js-private

No

**Testing**
Testing completed successfully in stage via:
- Compile individual destinations and loading them into ajs-next

This is part of rolling out ajs-next


.circleci/config.yml
```diff
@@ -1,113 +0,0 @@
...
-
-            #Determine range of SHAs for diffing files
-            COMMITS="${CIRCLE_SHA1}..${MASTER_COMMIT}"
-
-            #Check the diff on the commits
-            git diff --name-only $COMMITS | cut -d/ -f1 | sort -u > projects
-            echo -e "Modified directories:\n`cat projects`\n"
-
-            # If modified directories contain yarn, buildkite, or circleci directories
-            buildall=0
```
pooyaj, line 81: are we good to kill this logic? Do we have a similar step in buildkite?

nettofarah: circleCI doesn't pass anymore, regardless. We'd have to do the same thing in buildkite if we still care about this logic

pooyaj: I see, didn't notice that Circle is not doing anything before. All good then 👍

pooyaj (approved):

PR #555, opened 2021-02-03
[STRATCONN-588] Generate bundledConfigIds

gpsamson:
**What does this PR do?**
This pull request builds a list of `bundledds` based on the intersection between `bundled` and `maybeBundledConfigIds`. `bundledIds` is then appended to each event's metadata object.

The related ajs-renderer change can be found here: https://github.com/segmentio/ajs-renderer/pull/256.

**Example metadata object**
image
- `601c3fef8942cb2ad5d70cdf` is the FullStory instance that is not Connection Mode configurable because there is only a browser integration.
- `601c3c238942cb7cb5d70cde` is the Google Analytics instance configured in Device Mode.
- `601c4f8164b0701d3b6b78cd` is the HubSpot instance that is not Connection Mode configurable, but defaults to Device Mode because it's an a.js source.
- `601c40c46b6df7898d069abf` is the Amplitude instance that is configured explicitly in Cloud Mode.

No. We specifically are adding new metadata fields to avoid introducing breaking changes to `bundled` and `unbundled`.

**Testing**
- [x] Testing completed successfully in stage via end-to-end tests with the related `ajs-renderer` change.
- [x] Existing and new unit tests are passing

### DIFF
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -341,8 +341,7 @@ Segment.prototype.normalize = function(message) {
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
     msg._metadata.unbundled = this.options.unbundledIntegrations;
-    msg._metadata.bundledConfigIds = bundledConfigIds;
-    msg._metadata.unbundledConfigIds = this.options.unbundledConfigIds;
+    msg._metadata.bundledIds = bundledConfigIds;
   }
   this.debug('normalized %o', msg);
   this.ampId(ctx);
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -323,6 +323,9 @@ Segment.prototype.normalize = function(message) {
     var bundledConfigIds = []
     for (var i = 0; i < bundled.length; i++) {
       var name = bundled[i]
+      if (!maybeBundledConfigIds) {
+        break
+      }
       if (!maybeBundledConfigIds[name]) {
         continue
       }
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -340,8 +340,8 @@ Segment.prototype.normalize = function(message) {
 
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
-    msg._metadata.bundledConfigIds = bundledConfigIds;
     msg._metadata.unbundled = this.options.unbundledIntegrations;
+    msg._metadata.bundledConfigIds = bundledConfigIds;
     msg._metadata.unbundledConfigIds = this.options.unbundledConfigIds;
   }
   this.debug('normalized %o', msg);
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -18,7 +18,7 @@ var utm = require('@segment/utm-params');
 var uuid = require('@lukeed/uuid').v4;
 var Queue = require('@segment/localstorage-retry');
 
-const json = JSON;
+var json = JSON;
 
 /**
  * Cookie options
@@ -65,8 +65,8 @@ var Segment = (exports = module.exports = integration('Segment.io')
   .option('retryQueue', true)
   .option('addBundledMetadata', false)
   .option('unbundledIntegrations', []))
-  .option('unbundledIntegrationIds', [])
-  .option('maybeBundledConfigIds', []);
+  .option('unbundledConfigIds', [])
+  .option('maybeBundledConfigIds', {});
 
 /**
  * Get the store.
@@ -317,6 +317,7 @@ Segment.prototype.normalize = function(message) {
   }
   if (this.options.addBundledMetadata) {
     var bundled = keys(this.analytics.Integrations);
+    var maybeBundledConfigIds = this.options.maybeBundledConfigIds
 
     // Generate a list of bundled config IDs using the intersection of
     // bundled destination names and maybe bundled config IDs.
@@ -330,8 +331,8 @@ Segment.prototype.normalize = function(message) {
         continue
       }
 
-      for (var i = 0; i < maybeBundledConfigIds[name].length; i++) {
-        var id = maybeBundledConfigIds[name][i]
+      for (var j = 0; j < maybeBundledConfigIds[name].length; j++) {
+        var id = maybeBundledConfigIds[name][j]
         bundledConfigIds.push(id)
       }
     }
@@ -339,8 +340,8 @@ Segment.prototype.normalize = function(message) {
 
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
-    msg._metadata.unbundled = this.options.unbundledIntegrations;
     msg._metadata.bundledConfigIds = bundledConfigIds;
+    msg._metadata.unbundled = this.options.unbundledIntegrations;
     msg._metadata.unbundledConfigIds = this.options.unbundledConfigIds;
   }
   this.debug('normalized %o', msg);
--- a/integrations/segmentio/lib/index.js
+++ b/integrations/segmentio/lib/index.js
@@ -64,7 +64,9 @@ var Segment = (exports = module.exports = integration('Segment.io')
   .option('saveCrossDomainIdInLocalStorage', true)
   .option('retryQueue', true)
   .option('addBundledMetadata', false)
-  .option('unbundledIntegrations', []));
+  .option('unbundledIntegrations', []))
+  .option('unbundledIntegrationIds', [])
+  .option('maybeBundledConfigIds', []);
 
 /**
  * Get the store.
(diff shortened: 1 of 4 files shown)


integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
+      }
+    }
+
+
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
+    msg._metadata.bundledConfigIds = bundledConfigIds;
     msg._metadata.unbundled = this.options.unbundledIntegrations;
-    msg._metadata.bundledConfigIds = this.options.bundledConfigIds;
     msg._metadata.unbundledConfigIds = this.options.unbundledConfigIds;
```
gpsamson, line 345: I decided not to ignore empty `bundledConfigIds` and `unbundledConfigIds` arrays so it is clear when an event was tracked using this new version of `segmentio`.

integrations/segmentio/lib/index.js
```diff
@@ -18,7 +18,7 @@ var utm = require('@segment/utm-params');
 var uuid = require('@lukeed/uuid').v4;
 var Queue = require('@segment/localstorage-retry');
 
-const json = JSON;
+var json = JSON;
```
gpsamson, line 21: This unintentionally switched to `const` in #553.

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
+        var id = maybeBundledConfigIds[name][j]
+        bundledConfigIds.push(id)
+      }
+    }
+
+
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
+    msg._metadata.bundledConfigIds = bundledConfigIds;
     msg._metadata.unbundled = this.options.unbundledIntegrations;
```
mericsson, line 344:
```suggestion
	msg._metadata.unbundled = this.options.unbundledIntegrations;
    msg._metadata.bundledConfigIds = bundledConfigIds;
```

nit, I would keep order the same in case a system expects consistent json key order.

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
+
+    // Generate a list of bundled config IDs using the intersection of
+    // bundled destination names and maybe bundled config IDs.
+    var bundledConfigIds = []
+    for (var i = 0; i < bundled.length; i++) {
+      var name = bundled[i]
+      if (!maybeBundledConfigIds) {
+        break
+      }
+      if (!maybeBundledConfigIds[name]) {
```
mericsson, line 330: Looks like the casing is the same of destination names in `maybeBundledConfigIds` and `analytics.Integrations` e.g. `Google Analytics` === `Google Analytics` vs `google-analytics`.. but perhaps something to double check.

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
   }
   if (this.options.addBundledMetadata) {
     var bundled = keys(this.analytics.Integrations);
+    var maybeBundledConfigIds = this.options.maybeBundledConfigIds
+
+    // Generate a list of bundled config IDs using the intersection of
+    // bundled destination names and maybe bundled config IDs.
+    var bundledConfigIds = []
```
mericsson, line 324: I think `var bundledConfigIds = []` should be defined outside of this conditional for clarity. This works currently because of use of `var`.

integrations/segmentio/lib/index.js
gpsamson: Yup, the casing here is the same but they depend on two different sources.

    - `bundled` depends on the name that is defined within the integration code itself: https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/google-analytics/lib/index.js#L37
    - `bundledConfigIDs` depends on the defintiion creation name: https://github.com/segmentio/ajs-renderer/pull/256/files#diff-ca733fb7a822baffec24f4f426ccb07fce89f5b4d5856f848be4845d1d40e12bR44
    
    
The chance of these being different is unlikely but possible. Fwiw, the integration wouldn't receive the proper settings if the name within the integration code doesn't match the destination creation name because settings are keyed by the definition creation name.

integrations/segmentio/lib/index.js
gpsamson: In my stage tests, the order doesn't affect behavior.

integrations/segmentio/lib/index.js
gpsamson: Ahh. Good catch. 👍🏽

gpsamson: I don't think so -- `msg._metadata.bundledConfigIds` (L343) is defined within the conditional. I'm also mimicking the scope of `bundled`.

integrations/segmentio/lib/index.js
```diff
@@ -64,7 +64,9 @@ var Segment = (exports = module.exports = integration('Segment.io')
   .option('saveCrossDomainIdInLocalStorage', true)
   .option('retryQueue', true)
   .option('addBundledMetadata', false)
-  .option('unbundledIntegrations', []));
+  .option('unbundledIntegrations', []))
+  .option('unbundledConfigIds', [])
+  .option('maybeBundledConfigIds', {});
```
gpsamson, line 69: If we were to conditionally add these arrays in `ajs-renderer` we would still default to empty values here. 🤔

integrations/segmentio/lib/index.js
briemcnally: +1 to Gabe's comment, I spot checked a few destinations which I knew have undergone name changes and the name defined in integration code marched the definition creation name.

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
+      var name = bundled[i]
+      if (!maybeBundledConfigIds) {
+        break
+      }
+      if (!maybeBundledConfigIds[name]) {
+        continue
+      }
+
+      for (var j = 0; j < maybeBundledConfigIds[name].length; j++) {
+        var id = maybeBundledConfigIds[name][j]
```
briemcnally, line 335: Do we maybe want to see if `maybeBundledConfigIds[name]` has more than one configId? In example case where maybeBundledConfigIds looks like the below, wouldn't we want to collect stats since that would mean they have multiple device mode configs of a destination in device mode, which is not allowed?
```
maybeBundledConfigIds: {
       Fullstory: [ config1, config2]
}
```

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
+      }
+    }
+
+
     msg._metadata = msg._metadata || {};
     msg._metadata.bundled = bundled;
     msg._metadata.unbundled = this.options.unbundledIntegrations;
-    msg._metadata.bundledConfigIds = this.options.bundledConfigIds;
+    msg._metadata.bundledConfigIds = bundledConfigIds;
     msg._metadata.unbundledConfigIds = this.options.unbundledConfigIds;
```
gpsamson, line 345: `unbundledConfigIds` is not used in data plane so it could be removed later on.

integrations/segmentio/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@segment/analytics.js-integration-segmentio",
   "description": "The Segmentio analytics.js integration.",
-  "version": "4.3.1",
+  "version": "4.4.1",
```
juliofarah, line 4: nit: should this be 4.4.0 ?

gpsamson: Yup, thanks for catching,

briemcnally (approved):

integrations/segmentio/lib/index.js
mericsson: Makes sense. I agree.

integrations/segmentio/lib/index.js
mericsson: Added comment here regarding this https://github.com/segmentio/ajs-renderer/pull/256/files#r577105220

integrations/segmentio/lib/index.js
```diff
@@ -315,10 +317,31 @@ Segment.prototype.normalize = function(message) {
...
   if (this.options.addBundledMetadata) {
     var bundled = keys(this.analytics.Integrations);
+    var maybeBundledConfigIds = this.options.maybeBundledConfigIds
+
+    // Generate a list of bundled config IDs using the intersection of
+    // bundled destination names and maybe bundled config IDs.
+    var bundledConfigIds = []
+    for (var i = 0; i < bundled.length; i++) {
+      var name = bundled[i]
+      if (!maybeBundledConfigIds) {
```
mericsson, line 327: I think it would be clearer if this conditional was outside of the loop since it does not depend on anything in `bundled`.

mericsson (approved):
lgtm - just one code cleanliness issue https://github.com/segmentio/analytics.js-integrations/pull/555#discussion_r577106209

integrations/segmentio/lib/index.js
gpsamson: We'd have to refactor and wrap the for-loop in a conditional. I can go either way but was trying to avoid nested conditionals.

```
if (maybeBundledConfigIds) {
  for (var i = 0; i < bundled.length; i++) { ... }
}
```

mericsson (approved):

briemcnally (approved):

Issue #111, opened 2019-03-01
[appboy] integration snippet needs update (or replacement with npm require)


SegmentDestinationsBot:
_Migrated from [#33](https://github.com/segment-integrations/analytics.js-integration-appboy/issues/33) by @froodian_

The current integration is using a loading snippet at https://github.com/segment-integrations/analytics.js-integration-appboy/blob/fd6b5cdd65398c6cdaddb30d4f952e5a5de35384/lib/index.js#L124 that's somewhat different from the default one at https://github.com/Appboy/appboy-web-sdk#getting-started, and it's causing loading failures (js errors) in some environments.  The issue is that it explicitly gives appboy and appboyQueue the window namespace (window.appboy, window.appboyQueue), but pushes to the appboyQueue in the global namespace.  There were some updates on the Braze end around this snippet a while back to improve this situation, so it's possible it's simply outdated.

A larger point I want to raise here though is that this entire snippet with the appboyQueue structure was designed to be put in the `<head>` of an HTML page for naive clients in a raw html environment, and exists to handle method calls that happen before the script has loaded.  Given that in this integration, the script is instead loaded with `this.load('v2', function() {`, that snippet could be removed entirely and all methods on appboy simply called within the callback function.

Even simpler in fact, given the npm environment, would be to do `npm install --save-dev appboy-web-sdk`, `var appboy = require('appboy-web-sdk');` and then you can just immediately call `appboy.initialize` and start calling methods from there.  I'm not sure how you'd want to handle your v1 and v2 support however.

Given that there are a few different options here, I leave it to you to choose which one you want in your codebase, but that mismatch between `window.appboyQueue=` and `appboyQueue.push` should be rectified somehow, and I can tell you that internally the SDK draws from `appboyQueue` if it exists (and ignores it gracefully if it doesn't), so this code should either define and push to that, or omit it entirely.

Thanks!

PR #558, opened 2021-02-16
[braze fka appboy] update loading snippet for v2

davidbielik:
**What does this PR do?**
Fixes https://github.com/segmentio/analytics.js-integrations/issues/111

A race condition between `appboy.min.js` and the loading snippet causes an occasional error due to an old reference to a local `appboyQueue` variable. This replaces the v2 initialization with the latest v2 loading snippet to reference `window.appboyQueue` instead of `appboyQueue`

No

**Testing**
Testing completed successfully
image

n/a

n/a

no

n/a


mericsson:
Filed https://segment.atlassian.net/browse/DEST-3105 to track this on our end

davidbielik:
> Can you share what that error is?

Hey @mericsson sorry for the late reply. There was an update to our integration snippet which used a reference to a local variable `appboyQueue` instead of the `window.appboyQueue` reference. Depending timing and if a website used our API before the CDN file loaded, a client's integration may have tried to access an old reference to that variable.

Turns out it may have been a different cause for the customer (still investigating). We can close this for now and re-open if it pops up again. Thanks!

PR #561, opened 2021-02-23
Aptrinsic/region

nwolfe-xx:
**What does this PR do?**
Adds ability for customer to choose which region the PX subscription is hosted in.

No, it is backward compatible.  If data center is not specified, the historical behavior continues, the data center defaults to “united_states”.

**Testing**

Testing completed successfully: 

Updated unit tests to test new dataCenter option
Tested against production PX environment using the "Analytics.js-Integrations Tester" tool.  Verified three scenarios work as expected:
 - No value specified for dataCenter, defaults to "united_states"
 - Specified united_states
 - Specified eu

We are in the process of enabling an EU datacenter that will be used by customers who wish to have all of their data hosted in the EU.  We want to have the Segment code load a script from our EU datacenter instead of the current one from the US if specified by the customer.

We need to make a very similar change to the “Gainsight PX Cloud Server” integration.  The endpoint address needs to change based on the customer’s region selection (esp-eu.aptrinsic.com instead of esp.aptrinsic.com).  Note: The Cloud Server change is independent of the JS integration change.

Yes, we need a new “dataCenter” option.  It should be a dropdown/select with the following labels/values:
- United States/united_states
- EU/eu

The default option value for new integrations should be united_states, we do not need to force the user to make a selection.

Help Text below selector:
`For PX instances based in the European Union at https://app-eu.aptrinsic.com, select the EU option.  For all others (at https://app.aptrinsic.com), select the United States option.`

Mockup of config screen showing new region setting:
image

**Why is this PR raised?

We are hoping to allow our mutual customers to have the option to use the new EU datacenter.

**Are there any docs surrounding the change?

Once the change is live, we will upgrade the relevant screenshots and text on this page: https://support.gainsight.com/PX/Install_PX/02_Install_Gainsight_PX_using_Segment_for_web_apps

**What will the customer impact be as a result of the change?

Customers will be able to use Segment to load the PX tag from the EU data center and comply with their GDPR requirements.

**What will the customer impact be if the change is not implemented?

EU customers will need to load the PX tag manually (not use Segment)


danieljackins (approved):

danieljackins:
moving this over to https://github.com/segmentio/analytics.js-integrations/pull/574

Issue #562, opened 2021-02-24
@segmentio/send-json not being passed a timeout argument


texonidas:
While running segment locally in debug mode, I've run into an issue in lib/index.js on line 394. The send function being called (@segment/send-json) expects 5 arguments, (url, msg, headers, timeout, fn), however it is only being passed (url, msg, header, fn), and so I am getting constant "i is not a function" errors in my console.

iwakoscott:
I started seeing the same errors today as well 😅

shaunma:
It looks like in this recent commit https://github.com/segmentio/analytics.js-integrations/commit/cf751050547b0078836c01946c2c2d69f1e4e322#diff-ec90774f0561687711e2f7e544859dfcb2d4900f482bc5eab32fc644159503e1, `@segment/send-json` was upgraded from v3 to v4.

The [`json()`](https://github.com/segmentio/send-json/blob/3bb6f52056c9ab2414427959026bd1a23bce6c76/lib/index.js#L58) in v3 expects 4 arguments. The [same](https://github.com/segmentio/send-json/blob/4eb9bb097fb35f31c0475d2a9748b3d9433f6201/lib/index.js#L60) function in v4 now expects 5 arguments.

texonidas:
https://github.com/segmentio/analytics.js-integrations/pull/563

I've added pull request to fix this, I just used a 1s for send because I'm really not sure what was an appropriate value to use.

PR #563, opened 2021-02-24
add timeout of one second to enqueue non-retry send call

texonidas:
**What does this PR do?**
Fix a breaking bug from the RIP JSONv3 commit.

No, there's fixes.

**Testing**

No


texonidas:
https://github.com/segmentio/analytics.js-integrations/issues/562

PR #567, opened 2021-03-05
Better source maps

nettofarah:
**What does this PR do?**
Change the dynamic template scripts so that Source Maps are no longer shifted.
The previous implementation altered the number of lines in the original file, shifting every line by 2~3.
We no longer prepend lines to the file now though, which makes browsers happy again.

N/A

**Testing**
Testing completed successfully using local via:
- using the new `make local-server` command and loading all destinations into ajs-next


pooyaj (approved):
😮

PR #568, opened 2021-03-07
made functions async to reduce load time

rihanaster:
**What does this PR do?**
made all the function calls async to avoid any blocking calls and reduce website load time

No

**Testing**
Testing not required because - functions are just made async, there is no significant change in the code.

NA

NA

No

NA


pooyaj:
@rihanaster I don't think making these functions `async` improves the performance. The body of the function still executes synchronously as the function implementation is not asynchronous.

PR #569, opened 2021-03-08
Add turbo for lint, test

jaredpalmer:
**What does this PR do?**
Switch from `lerna` to `turbo` for employees / CI, as requested by @nettofarah 

- [x] Lint
- [x] Test
- [ ] Build
- [ ] CI/CD caching

No.

**Testing**

Turborepo is new build system for monorepos. Among other features, `turbo` provides Bazel-like remote build caching and task running, but in a way that is nearly fully compatible with Lerna. 

`turbo` is installed as an optional dependency since it is not open source and requires an access token to download from the Turborepo registry (npm.turborepo.com). An `.npmrc` is added to address this. In addition, Lerna commands we left runnable for open source contributors.

N/A

No

https://turborepo.com/docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ package-lock.json
 yarn-error.log
 dist/
 build/
+.npmrc
+turbo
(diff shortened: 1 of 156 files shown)


nettofarah:
This is looking great so far, @jaredpalmer. Thanks for taking a stab at it.

PR #570, opened 2021-03-09
Google Analytics 4

gpsamson:
**What does this PR do?**
This pull request introduces a new Google Analytics 4 to service the new GA4 property type which varies greatly from Universal property types. It includes functionality up to P2 in the work breakdown.

The complete spec can be found in the [GA4 A.js Destination SDD](https://paper.dropbox.com/doc/SDD-GA4-A.js-Destination--BC~O41uLfl9Hnz5VkbFnN9gWAg-eVNt15tiLlZjZotG3lkKk).

No

**Testing**
- [x] Testing completed successfully in stage by manually sending events.
- [x] Unit tests complete

Detailed testing can be found here: https://paper.dropbox.com/doc/GA4-A.js-Test-Plan--BHKEytO0krCgb7Rq0ETLY6TBAg-hIeXHKehkOvlhqONgOGdV

A completely new definition can be found in Partner Portal for `google-analytics-4`

https://developers.google.com/analytics/devguides/collection/ga4/tag-guide is heavily referenced throughout the code.


integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,310 @@
...
+    for (var i = 0; i < measurementIds.length; i++) {
+      config.push([
+        'config',
+        measurementIds[i],
+        {
+          send_page_view: false
+        }
+      ]);
+    }
+  }
```
briemcnally, line 79: It looks like we add the GA properties then if they have another setting enabled ie `disablePageViewMeasurement` push the same properties to the config with with `send_page_view: false` Curious if this causes any dupes? Looks like config is just an array

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Group
+ *
+ * @api public
+ * @param {Facade.Group} group
+ */
+GA4.prototype.group = function(group) {
+  var props = reject({
+    group_id: group.groupId()
+  });
```
briemcnally, line 234: if this can be empty from the reject if there is no group_id() should we not log this event?

brennan: I think `groupId` is always required for `group` events, [seen here](https://github.com/segmentio/loosely-validate-event/blob/8b247ad20f817645d569d4e664ff372dcc3be129/index.js#L51), so we probably don't need to wrap the props object with `reject`.

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,310 @@
...
+  if (!this.options.disablePageViewMeasurement) {
+    return;
+  }
+
+  // Track categorized pages setting needed here?
+
+  var props = page.properties();
+  var name = page.fullName();
+
+  // var language = ;
```
brennan, line 257: Not sure if we have or are working on a plan for language? I see in the SDD that Segment doesn't have it specced for a.js events.

integrations/google-analytics-4/lib/index.js
briemcnally: SG. +1 to removing reject

integrations/google-analytics-4/lib/index.js
gpsamson: Not sure if this would cause dupes. I think it would be cleaner to push the properties once, so I can make that change.

integrations/google-analytics-4/lib/index.js
gpsamson: I don't think it is part of the Segment spec which is why I left it commented. It is mapped in the SDD but I don't think it's crucial for the pilot version. I can remove for now and re-add if there are asks for it!

integrations/google-analytics-4/lib/index.js
gpsamson: Thank you so much for calling this out! Seems when we call `config` multiple time may overwrite others 🤔.

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,275 @@
...
+    cookie_prefix: opts.cookiePrefix,
+
+    /**
+     * Cookie Expiration
+     * https://developers.google.com/analytics/devguides/collection/ga4/cookies-user-id#cookie_expiration
+     */
+    cookie_expires: opts.cookieExpiration,
+  };
+
+  var sets = [
```
briemcnally, line 96: Should this be an array of objects? A little bit confused by the different format of Cookie Flags and the rest of them. Could be missing something

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,275 @@
...
+
+    userProperties[userProp] = value;
+  }
+
+    /**
+     * Map the user_id property the Send User ID setting is enabled. Note that the user ID
+     * can be appended as part of the user_properties object instead of being configured by
+     * an explicit command.
+     * https://developers.google.com/analytics/devguides/collection/ga4/cookies-user-id#set_user_id
+     */
```
briemcnally, line 182: Small nit: should we move the comment to L to be aligned

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,275 @@
...
+ * @api public
+ * @param {Track} track
+ */
+
+GA4.prototype.track = function(track) {
+
+  var mappings = this.options.customEventsAndParameters;
+
+  for (var i = 0; i < mappings.length; i++) {
+    var mapping = mappings[i]; // Type check for object?
```
briemcnally, line 246: Based off the format for the mixed field setting for this, would it ever be possible for it to be anything other than object? If not, probably don't need type check.

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,275 @@
...
+ *
+ * @api public
+ * @param {Track} track
+ */
+
+GA4.prototype.track = function(track) {
+
+  var mappings = this.options.customEventsAndParameters;
+
+  for (var i = 0; i < mappings.length; i++) {
```
briemcnally, line 245: Small nit: Sometimes it is hard to comprehend the Mixed Setting type. I know for AA we include a code comment re: format( [seen here](https://github.com/segmentio/analytics.js-integrations/blob/ede07d097a7b447a58e2f08cbb2c0eefbaf828cb/integrations/adobe-analytics/lib/index.js#L712-L730) ). Might be helpful to include one here for other contributors less familiar with this settings ie. 
```
{
  segmentEvent: 'stringSegmentEvent',
  googleEvent: 'stringGoogleEvent',
  parameters: { segmentParam: 'googleParam'}
}
```

briemcnally: Not sure if ☝️ is even the right format, based off below looks like parameters is an array of objects?

integrations/google-analytics-4/lib/index.js
gpsamson: gtag.js has a [`set` command](https://developers.google.com/gtagjs/reference/api#set) and it's really confusing. Sometimes it accepts a [single object argument](https://developers.google.com/analytics/devguides/collection/ga4/cookies-user-id#cookie_flags) (L1) and sometimes it accepts [multiple arguments](https://developers.google.com/analytics/devguides/collection/ga4/display-features#disable_all_advertising_features) (L2):
```
window.gtag('set', { cookie_flags: opts.cookieFlags })
window.gtag('set', 'allow_google_signals', !opts.disableAllAdvertisingFeatures)
```

The `sets` var we have is an array of arrays to represent a list of set commands and the arguments to pass.

integrations/google-analytics-4/lib/index.js
gpsamson: Good point. I'll add an example 🙌🏽

integrations/google-analytics-4/lib/index.js
briemcnally: Gotcha makes sense. Thanks for clarifying

gpsamson:
@brennan @briemcnally - Jen and I went through the test plans on Friday and all functionality is working as expected. There may be changes to the setting names today, but not a drastic change in behavior.

Thanks for the reviews so far!

briemcnally (approved):
LGTM! Nice work @gpsamson

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,303 @@
...
+ * @param {Facade.Page} page
+ */
+GA4.prototype.page = function(page) {
+  // If the Send Google's Automatic Page View Measurement setting is set to true then
+  // don't handle page calls to avoid duplicate page_view events.
+  if (this.options.sendAutomaticPageViewEvent) {
+    return;
+  }
+
+  // Track categorized pages setting needed here?
```
brennan, line 246: Have we been able to look into this? :)

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,303 @@
...
+ * @api public
+ * @param {Track} track
+ */
+
+GA4.prototype.track = function(track) {
+
+  var mappings = this.options.customEventsAndParameters;
+
+  for (var i = 0; i < mappings.length; i++) {
+    var mapping = mappings[i]; // Type check for object?
```
brennan, line 274: Wondering per the comment if we should put a type check for object here?

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,303 @@
...
+    var segmentEvent = mapping.segmentEvent;
+    var googleEvent = mapping.googleEvent;
+  
+    if (!segmentEvent || !googleEvent || segmentEvent !== track.event()) {
+      continue;
+    }
+
+    var parameterMappings = mapping.parameters || []; // Type check for array?
```
brennan, line 282: Also curious about this comment re: a type check for array. Is that something we should add?

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,303 @@
...
+    var parameters = {};
+
+    // Map Segment event fields to Google Event Parameters.
+    // Text map settings that are nested in a mixed settings take on a different shape
+    // than a top-level text map setting.
+    // eg; [{ key: 'properties.genre', value: 'primary_genre }]
+    //
+    for (var j = 0; j < parameterMappings.length; j++) {
+      var map = parameterMappings[j] || {}; // Type check for object?
```
brennan, line 291: Another type-check comment here as well.

integrations/google-analytics-4/lib/index.js
```diff
@@ -0,0 +1,303 @@
...
+      if (!map.key || !map.value) {
+        continue;
+      }
+
+      var param = map.value;
+      var value = track.proxy(map.key);
+      parameters[param] = value;
+    }
+
+    window.gtag('event', googleEvent, reject(parameters));
```
brennan, line 301: Wondering if we need `reject` here given line 292 above - if `map. value` is falsy, then we don't add the key:value pair to the parameters map in the first place.

gpsamson: Good call - it's pretty redundant!

integrations/google-analytics-4/lib/index.js
gpsamson: I ran it by Jen but no response yet. I think we can leave it out for now and add it if customers request it. What do you think?

gpsamson: Jen recommended we omit it until this functionality is requested.

integrations/google-analytics-4/lib/index.js
brennan: @gpsamson Not a big deal, but looks like we still have a type-check comment here.

brennan (approved):

integrations/google-analytics-4/lib/index.js
gpsamson: Oops, thx! I will remove it prior to merging.

gpsamson:
Thanks for the review @brennan @briemcnally :tribe:

I downgraded the version to `0.0.1` since this will be in private beta for a few weeks and we may make changes that break existing behavior (eg; default mappings).

dobesv:
Is there any process or timeline for this to be published to npm?

PR #572, opened 2021-03-12
Added statusCode check for NetworkErrors in Segmentio integration

Dahaden:
**What does this PR do?**

Changes Segment.io integration to retry events when there is a NetworkError from the browser.

 No

**Testing**

Added test case.

Noticed this the other day when debugging issues in the browser (IE, turning off WiFi, or telling the tab to go into offline mode)

I dont believe so

No

https://fetch.spec.whatwg.org/#concept-network-error
Which is referenced from https://xhr.spec.whatwg.org/#response

> A network error is a response whose status is always 0, status message is always the empty byte sequence, header list is always empty, and body is always null.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6526,7 +6526,7 @@ extend@3.0.1:
   resolved "https://registry.yarnpkg.com/extend/-/extend-3.0.1.tgz#a755ea7bc1adfcc5a31ce7e762dbaadc5e636444"
   integrity sha1-p1Xqe8Gt/MWjHOfnYtuq3F5jZEQ=
 
-extend@3.0.2, extend@^3.0.0, extend@^3.0.1, extend@^3.0.2, extend@~3.0.2:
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -6526,7 +6526,7 @@ extend@3.0.1:
   resolved "https://registry.yarnpkg.com/extend/-/extend-3.0.1.tgz#a755ea7bc1adfcc5a31ce7e762dbaadc5e636444"
   integrity sha1-p1Xqe8Gt/MWjHOfnYtuq3F5jZEQ=
 
-extend@3.0.2, extend@^3.0.0, extend@^3.0.1, extend@^3.0.2, extend@~3.0.2:
+extend@3.0.2, extend@^3.0.0, extend@^3.0.2, extend@~3.0.2:
```
Dahaden, line 6529: Sorry didnt notice this change when committing. It was applied automatically by `yarn`.
Happy to revert :)

PR #575, opened 2021-03-24
Fix GA4 set bug

gpsamson:
**What does this PR do?**
This pull request fixes a bug where GA4 `set` commands were not being passed to gtag.js with the `"set"` string. This wasn't caught when loaded locally because GA will swallow the errors.

N/A

**Testing**
- [x] Added unit tests to ensure `"set"` is passed
- [x] Testing completed successfully using a stage snippet on a non-local site


integrations/google-analytics-4/test/index.js
```diff
@@ -145,20 +145,21 @@ describe('Google Analytics 4', function () {
             analytics.equal(window.ga4DataLayer[2][2]['cookie_expires'], 21)
 
             // cookie_flags uses the `set` command
-            analytics.equal(window.ga4DataLayer[3][0]['cookie_flags'], 'SameSite=None;Secure')
+            analytics.equal(window.ga4DataLayer[3][0], 'set')
```
brennan, line 148: Should we use a `deepEqual` here like the other updated tests!

brennan (approved):

integrations/google-analytics-4/test/index.js
gpsamson: Yeah, that may be cleaner. I'll change before I merge!

brennan: Nice! Definitely a nit - but it does look nice this way! Thank you!

PR #597, opened 2021-05-11
Update Nielsen integration mappings and add new setting to Nielsen DCR

kdaswani:
**What does this PR do?**
This PR updates our Nielsen DCR and Nielsen DTVR (to be added) integrations to meet Nielsen requirements. Changes requested by Nielsen are outlined in [this spreadsheet](https://docs.google.com/spreadsheets/d/192EYcjqiPBdTEKDCKQ0iJB6gjJH3CDQDeyenYmuAe-k/edit#gid=0).

- [JIRA](https://segment.atlassian.net/browse/LIBMOBILE-369)

No.

**Testing**
Testing in progress

Fox has been trying to get certified by Nielsen for a year+ now. In order to get certified using the Segment-Nielsen integrations, our integrations to meet Nielsen's requirements. This should be the last set of changes blocking certification.

Yes. The same changes have already been made to iOS (published new versions) and Android (being tested by customer before publishing new versions):
- iOS DCR: https://github.com/segment-integrations/analytics-ios-integration-nielsen-dcr/commit/010dc9779bbcbfdb6199e93f3fd1fff0321d9b91
- iOS DTVR: https://github.com/segment-integrations/analytics-ios-integration-nielsen-dtvr/commit/06eb3a23d2c781d8eeb311d3e220bcefc7c1270a
- Android DCR: https://github.com/segment-integrations/analytics-android-integration-nielsen-dcr/commit/f059bfb5259b7acaa2df3b1a98927a3fb86d7ff5
- Android DTVR: https://github.com/segment-integrations/analytics-ios-integration-nielsen-dtvr/commit/06eb3a23d2c781d8eeb311d3e220bcefc7c1270a

Yes. The Nielsen DCR integration requires a new setting for `customSectionProperty`. This setting is already live in prod (Partner Portal) as it is used in iOS and Android too. The setting allows a customer to choose which property on their page calls maps to the Nielsen `section` field. If nothing is set, we will fallback on the page name which is the existing behavior.

See spreadsheet linked above.

### DIFF
--- a/integrations/nielsen-dcr/lib/index.js
+++ b/integrations/nielsen-dcr/lib/index.js
@@ -22,6 +22,7 @@ var NielsenDCR = (module.exports = integration('Nielsen DCR')
   .option('adAssetIdPropertyName', '')
   .option('subbrandPropertyName', '')
   .option('clientIdPropertyName', '')
+  .option('customSectionProperty', '')
   .option('sendCurrentTimeLivestream', false)
   .option('contentLengthPropertyName', 'total_length')
   .option('optout', false)
@@ -100,10 +101,21 @@ NielsenDCR.prototype.loaded = function() {
 
 NielsenDCR.prototype.page = function(page) {
   var integrationOpts = page.options(this.name);
+  var customSectionName;
+
+  //Allow customer to pick property to source section from otherwise fallback on page name
+  if (this.options.customSectionProperty) {
+    customSectionName = page.proxy(
+      'properties.' + this.options.customSectionProperty
+    );
+  }
+  var defaultSectionName = page.fullName() || page.name() || page.event();
+  var sectionName = customSectionName || defaultSectionName;
+
   var staticMetadata = reject({
     type: 'static',
     assetid: page.url(), // *DYNAMIC METADATA*: unique ID for each article **REQUIRED**
-    section: page.fullName() || page.name() || page.event(), // *DYNAMIC METADATA*: section of site **REQUIRED**
+    section: sectionName, // *DYNAMIC METADATA*: section of site **REQUIRED**
     segA: integrationOpts.segA, // *DYNAMIC METADATA*: custom segment
     segB: integrationOpts.segB, // *DYNAMIC METADATA*: custom segment
     segC: integrationOpts.segC // *DYNAMIC METADATA*: custom segment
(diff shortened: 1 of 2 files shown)


kdaswani:
Confirmation from Nielsen contract on how to handle DTVR completion and interruption events as their public documentation is incorrect:
Screen Shot 2021-05-07 at 11 47 55 AM

kdaswani:
Hi @bsneed! I have made changes to our Nielsen DCR and DTVR web integrations, as requested by Nielsen. The changes made are outlined in the linked spreadsheet and JIRA. I've already made similar changes on iOS and Android so this is the last platform we need to tackle. I added some new unit tests and locally tested, and everything looks good. Would you or someone on the Libraries team be able to review and approve my changes? I'm newer to releasing web device-mode changes, but in the past we have pinned Fox's staging source(s) to a version of a.js with the changes so that they could test before we add to a release train. Is this possible? @brennan has helped with this in the past. thanks!

kdaswani:
Source IDs to pin:

pooyaj:
hey @kdaswani, I'll take a look shortly

kdaswani:
Hi @pooyaj have you had a chance to take a look at this PR? Thanks!

yarn.lock
```diff
@@ -2726,9 +2726,9 @@ analytics-events@^1.2.0:
   integrity sha1-sDCnqCv+tF7w2qdNIM+ai6TUvcs=
 
 analytics-events@^2.0.2, analytics-events@^2.1.0:
-  version "2.2.0"
```
pooyaj, line 2729: @kdaswani is this change intentional? curious if this will impact other destinations as well.

pooyaj (commented):
Just one concern around the changes in the yarn.lock. otherwise lgtm

pooyaj (approved):
After comparing the `segment-events` 2.2.0 vs. 2.2.5, the changes are non-breaking so this should be safe.

yarn.lock
kdaswani: yes! I had to add a new event for `Video Playback Exited` to analytics-events so we needed to update the version here.

kdaswani:
Added a small change to fix a bug with how `length` is populated. We weren't previously checking if the setting was undefined/empty, which would cause the integration to put the entire `content` object in the `length` field for ad events and never move to the `else` statement. I.e.
Screen Shot 2021-06-29 at 11 13 09 AM

By adding a check for if the setting exists, we move to the `else` statement if it is empty and `length` is populated correctly:
Screen Shot 2021-06-29 at 11 15 59 AM

Confirmed all unit tests still pass:
Screen Shot 2021-06-29 at 1 51 51 PM

@pooyaj can you take a look? I think I'll need to update the SHA in the beta version for Nielsen DCR here: https://github.com/segmentio/analytics.js-private/pull/632/files. And then pin the new commit to the three dev sites previously pinned. Is that correct? Thanks!

kdaswani:
Updated mapping for `assetid` for StaticStart content to be a SHA256 hash of url. This is because Nielsen does not allow special characters in the `assetid` and URLs can sometimes contain prohibited characters. They require `assetid` is unique for each page/article, but it does not need to be descriptive of the actual page. They have confirmed that a deterministic hash of URL (i.e. SHA256) will work. Updated unit tests and all passed. Tested locally using compiler and `assetid` is correctly set as the SHA256 hash of http://localhost:3000/.
Screen Shot 2021-07-06 at 12 48 31 PM

PR #601, opened 2021-05-12
Fix undefined utm_ params

juliofarah:
**What does this PR do?**

This PR changes the `utm` params parsing to guard against undefined values and empty keys, preventing cases like
```
- https:/localhost:3000/?utm_
- https://niallzato.github.io/?utm_foo=
```
from messing with tracking.

https://segment.atlassian.net/browse/LIBWEB-667

**Testing**
- Testing completed successfully with unit tests and by changing the compiled code directly into the customer's website:
image
image

### DIFF
--- a/integrations/segmentio/lib/utm.js
+++ b/integrations/segmentio/lib/utm.js
@@ -10,9 +10,9 @@ function utm(query) {
   var acc = {}
   for (var i = 0; i < split.length; i++) {
     var k = split[i].split('=')[0]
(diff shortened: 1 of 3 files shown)


danieljackins (approved):
LGTM! Can we add a unit test?

pooyaj (approved):

juliofarah:
> LGTM! Can we add a unit test?

Sure thing! this is still just a draft to validate with the team if these are really the changes we want

PR #606, opened 2021-06-01
[Customer.io] Allow for a separate script to be loaded for EU

juliofarah:
**What does this PR do?**
This PR adds the ability to choose between the global or EU datacenters for Customer.io.
Segment's app's UI will be updated to reflect such option once this PR lands - the option has been implemented on the backend, but it's invisible given it doesn't work on the library side yet.

**Testing**
- Testing completed successfully using Analytics 2.0 locally and a bundled version of Customer.io, uploaded to stage's CDN
image


pooyaj (approved):
lfg :shipit:

PR #607, opened 2021-06-01
Adobe analytics chromecast option

Isuru-F:
NOTE: 
This isn't yet ready for merging - and should remain unmerged in its current state due to the option.chromecastMode being set to true in this branch.

This PR is to get feedback required from the eng team

**What does this PR do?**

- Implement the Chromecast adobe analytics SDK
- Mapped the standard video event spec events to the Chromecast sdk video events
- Implemented unit tests

Setting the chromecast option to true - this is for this branch only required for a specific customer

**Testing**
Unit tests - for Chromecast SDK created in new unit test file chromecast.index.test

All PRs must follow the process for change control as outlined in:
https://segment.atlassian.net/wiki/spaces/GRC/pages/453935287/Reinforcing+Change+Control

- Testing completed successfully using <how did you test, environment>; or
- Testing not required because <explain why you think testing isn't needed>

--->

Eventually yes - but for now it will be hardcoded in this branch with chromecast set true

https://paper.dropbox.com/doc/Google-PRD-SDD-Segment-Analytics.js-Adobe-Chromecast-SDK-Integration--BLb8LAr0IRU_6dyQ_AegXuGpAg-x3NoHj9IM0QTrH7zMDiWo#:uid=083588252113878478874898&h2=Event-Mappings


integrations/adobe-analytics/test/chromecast.index.test.js
```diff
@@ -0,0 +1,491 @@
...
+  var analytics;
+  var adobeAnalytics;
+  var user;
+  var options = {
+    version: 2,
+    chromecastToggle: true,
+    reportSuiteId: 'sgmtest',
+    trackingServerUrl: 'exchangepartnersegment.sc.omtrdc.net',
+    trackingServerSecureUrl: '',
+    heartbeatTrackingServerUrl: 'https://exchangepartnersegment.hb.omtrdc.net',
```
kdaswani, line 20: update to `heartbeatTrackingServerUrl: 'https://exchangepartnersegment.hb-api.omtrdc.net',` for tests to work

kdaswani: @Isuru-F `index.test.js` unit tests are failing, but I'm not sure those matter since they are for the JS (non-Chromecast) SDK?

integrations/adobe-analytics/lib/index.js
```diff
@@ -139,10 +146,83 @@ AdobeAnalytics.prototype.initialize = function() {
...
+
+    this.load('chromecast', function () {
+      self.ready();
+    });
+  }
   // Load the larger Heartbeat script only if the customer has it enabled in settings.
   // This file is considerably bigger, so this check is necessary.
-  if (options.heartbeatTrackingServerUrl) {
-    this.load('heartbeat', function() {
+  if (!options.chromecastMode && options.heartbeatTrackingServerUrl) {
```
mericsson, line 224: Prefer if we did not add `!options.chromecastMode` to this conditional and next.

Since this `AdobeAnalytics.prototype.initialize` is quite large I propose we break out chromecast initialization separately.

```js
AdobeAnalytics.prototype.initialize = function () {
// initial setup
if (options.chromecastMode) {
  this.initializeChromeCastMode()
} else if (options.heartbeatTrackingServerUrl) {
  // existing heartbeat init
} else {
  // existing default init
}
```

integrations/adobe-analytics/lib/index.js
```diff
@@ -46,7 +47,7 @@ var AdobeAnalytics = (module.exports = integration('Adobe Analytics'));
  * @param {*} value
  */
 
-AdobeAnalytics.sOption = function(field, value) {
+AdobeAnalytics.sOption = function (field, value) {
```
mericsson, line 50: Looks like there are a lot of diffs resulting from this styling change. Is it something auto-generated by your IDE? It would help to not have these diffs to reduce noise in the PR.

integrations/adobe-analytics/lib/index.js
```diff
@@ -125,10 +129,13 @@ AdobeAnalytics.global('s')
  * @api public
  */
 
-AdobeAnalytics.prototype.initialize = function() {
+AdobeAnalytics.prototype.initialize = function () {
   var options = this.options;
   var self = this;
 
+
+  options.chromecastMode = true; //Hard coding this to true for this branch, will be switched to a flag/setting if merged into main
```
mericsson, line 137: Instead of hard-coding which creates noise for all the tests, can we ask the customer to update their implementation to include this setting using the integrations object?

https://segment.com/docs/connections/spec/common/#integrations

I believe you can do something like

```json
"integrations": {
  "Adobe Analytics": {
    "chromecastMode": true
  }
}

File: integrations/adobe-analytics/lib/index.js [hunk 5]
```diff
@@ -216,14 +298,117 @@ AdobeAnalytics.prototype.initialize = function() {
     });
   }
 };
+function chromecastHeartbeatVideoStart(track) {
```
mericsson, line 301: Let's put all of the heartbeat functions in a separate file to try to keep this lines of code in this file more manageable.

e.g. a file named `heartbeat.js` can have all of these functions `export`ed and then have this `lib/index.js` import and call them appropriately.

integrations/adobe-analytics/lib/index.js
```diff
@@ -216,14 +298,117 @@ AdobeAnalytics.prototype.initialize = function() {
...
+};
 
 /**
  * Adobe Analytics is loaded if the `window.s_gi` function exists.
  *
  * @return {Boolean} loaded
  */
 
-AdobeAnalytics.prototype.loaded = function() {
+
```
mericsson, line 410: extra newline
```suggestion
```

integrations/adobe-analytics/lib/index.js
```diff
@@ -234,56 +419,74 @@ AdobeAnalytics.prototype.loaded = function() {
...
+          window.s.visitorID = userId;
+      }
     }
-  }
 
-  // Attach some variables on the `window.s` to be sent with the call
-  update(pageName, 'events');
-  updateCommonVariables(page, this.options);
+    // Attach some variables on the `window.s` to be sent with the call
+    update(pageName, 'events');
```
mericsson, line 472: more diffs from style changes -- we should avoid creating this diff to make PR more focused on functional changes.

integrations/adobe-analytics/lib/index.js
```diff
@@ -293,12 +496,13 @@ AdobeAnalytics.prototype.page = function(page) {
...
  */
 
-AdobeAnalytics.prototype.track = function(track) {
+AdobeAnalytics.prototype.track = function (track) {
   // Delete any existing keys on window.s from previous call
   clearKeys(dynamicKeys);
 
   var eventName = track.event().toLowerCase();
 
+  console.log(eventName)
```
mericsson, line 505:
```suggestion
```

integrations/adobe-analytics/lib/index.js
mericsson, line 472: more diffs from style changes -- we should avoid creating this diff to make PR more focused on functional changes.

integrations/adobe-analytics/lib/index.js
```diff
@@ -309,6 +513,7 @@ AdobeAnalytics.prototype.track = function(track) {
   }
   // Check if Segment event is mapped in settings; if not, noop
   var isMapped = this.isMapped(eventName);
+  console.log(isMapped)
```
mericsson, line 516:
```suggestion
```

integrations/adobe-analytics/lib/index.js
```diff
@@ -418,50 +623,78 @@ AdobeAnalytics.prototype.checkoutStarted = function(track) {
  * @param {string} adobeEvent Adobe standard event
  */
 
-AdobeAnalytics.prototype.processEvent = function(msg, adobeEvent) {
-  var merchEvents = getMerchConfig(msg, this.options);
+AdobeAnalytics.prototype.processEvent = function (msg, adobeEvent) {
   var properties = msg.properties();
+  let adobeEvents = [];
```
mericsson, line 628:
```suggestion
  var adobeEvents = [];
```
browser compatibility

integrations/adobe-analytics/lib/index.js
Isuru-F: @mericsson is this just for the chromecast hb events or also including the original hb events?

Isuru-F: I've moved out the chromecast hb events to a new file, I feel a little uncomfortable making such a dramatic change to the existing hb events, but let me know if that's what is best here.

.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
mericsson, line 2: We should not check in any `.vscode` configs to this repo.

integrations/adobe-analytics/lib/heartbeat.js
```diff
@@ -0,0 +1,180 @@
+
+var heartbeats = {
+
+  extractMediaMetadata: function (track) {
+    var props = track.properties();
+    var mediaMetadata = {};
+    debugger
```
mericsson, line 7:
```suggestion
```

integrations/adobe-analytics/lib/heartbeat.js
```diff
@@ -0,0 +1,180 @@
+
+var heartbeats = {
+
+  extractMediaMetadata: function (track) {
```
mericsson, line 4: Please add [jsdoc](https://jsdoc.app/tags-returns.html) to document:
- `@returns`
- `@param`
- brief summary of function intent

integrations/adobe-analytics/lib/heartbeat.js
```diff
@@ -0,0 +1,180 @@
...
+    for (var prop in segAdbMap) {
+      stdVidMeta[segAdbMap[prop]] = props[prop] || 'no ' + segAdbMap[prop];
+    }
+    mediaInfo[standardMediaMetadata] = stdVidMeta;
+
+    window.ADBMobile.media.trackSessionStart(mediaInfo, mediaMetadata);
+  },
+
+
+
```
mericsson, line 80:
```suggestion
```

inconsistent spacing

integrations/adobe-analytics/lib/heartbeat.js
```diff
@@ -0,0 +1,180 @@
...
+    window.ADBMobile.media.trackEvent(window.ADBMobile.media.Event.SeekComplete, null, mediaMetadata);
+  },
+  chromecastBufferStarted: function (track) {
+    var mediaMetadata = window.extractMediaMetadata(track); 
+    window.ADBMobile.media.trackEvent(window.ADBMobile.media.Event.BufferStart, null, mediaMetadata);
+  },
+  chromecastQualityUpdated: function (track) {
+    var mediaMetadata = window.extractMediaMetadata(track); 
+    var props = track.properties();
+    let qosInfo = {
```
mericsson, line 156:
```suggestion
    var qosInfo = {
```
I believe we cannot use `let` yet in a.js 1.0.

integrations/adobe-analytics/lib/heartbeat.js
mericsson: Probably unnecessary for a lot of `chromecast...` functions e.g. `chromecastHeartbeatVideoStart` but would be nice on the more custom / less self-explanatory functions.

integrations/adobe-analytics/lib/index.js
```diff
@@ -10,7 +10,8 @@ var each = require('@ndhoule/each');
 var iso = require('@segment/to-iso-string');
 var Track = require('segmentio-facade').Track;
 var trample = require('@segment/trample');
-
+const analytics = require('@segment/analytics.js-core');
```
mericsson, line 13:
```suggestion
var analytics = require('@segment/analytics.js-core');
```

`const` not yet available in analytics.js 1.0.

integrations/adobe-analytics/lib/index.js
```diff
@@ -46,7 +47,7 @@ var AdobeAnalytics = (module.exports = integration('Adobe Analytics'));
  * @param {*} value
  */
 
-AdobeAnalytics.sOption = function(field, value) {
+ AdobeAnalytics.sOption = function(field, value) {
```
mericsson, line 50:
```suggestion
AdobeAnalytics.sOption = function(field, value) {
```

integrations/adobe-analytics/test/chromecast.index.test.js
```diff
@@ -0,0 +1,528 @@
...
+
+    });
+    describe('#track', function () {
+      let bufferStartSub;
+      beforeEach(function () {
+        analytics.stub(window.ADBMobile.analytics, 'trackAction');
+        analytics.stub(window.ADBMobile.analytics, 'createMediaObject');
+
+
+
```
mericsson, line 116: inconsistent spacing

integrations/adobe-analytics/lib/heartbeat.js
```diff
@@ -0,0 +1,180 @@
+
+var heartbeats = {
```
mericsson, line 2: Thanks for breaking this out! I think this is specific to chromecast heartbeat which makes sense. I think we can make that more explicit in the name of the object and filename.

```suggestion
var chromecastHeartbeatUtil = {
```

integrations/adobe-analytics/lib/index.js
Isuru-F: this indentation is correct as it's wrapped in an if now, please re-open if you disagree.

integrations/adobe-analytics/lib/heartbeat.js
Isuru-F: Added jsdoc for the only helper method in that file extractMediaMetadata

integrations/adobe-analytics/lib/index.js
Isuru-F: @kdaswani  - I'll leave this one hardcoded and unresolved and let you and @mericsson discuss/update

integrations/adobe-analytics/lib/chromecast-heartbeat.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+    return mediaMetadata;
+  },
+  
+  chromecastInit: function (track) {    
+    var props = track.properties();
+    var mediaMetadata = window.extractMediaMetadata(track);   
+    window.ADBMobile.config.setDebugLogging(true);
+    var qosInfoSettings = {
+      bitrate: props.bitrate | 1,
```
mericsson, line 27: Should this be logical OR `||` rather than bitwise OR? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR

Isuru-F: great catch - it should be OR.

integrations/adobe-analytics/lib/index.js
kdaswani: discussed offline with Marcus and we will keep this hardcoded for now. We will pin a version of a.js with this code to Fox's Chromecast sources for testing. In the short/medium-term, we plan to roll these changes into the existing Adobe integration and adding the boolean setting to partner portal, at which point we'll update this to reference the setting (instead of hardcoding).

integrations/adobe-analytics/lib/index.js
kdaswani: i will fix this along with any other style changes!

integrations/adobe-analytics/lib/index.js
kdaswani: will update this to swap the conditional!

integrations/adobe-analytics/lib/index.js
kdaswani: I think this specific change will have a style diff since it needs to be wrapped in an `else` statement whereas before it wasn't, so it's indented further in now

kdaswani (commented):
@mericsson I cleaned up the formatting and incorporated remaining feedback as best I could! There are a couple of places where the indentation changes are necessary since we wrap the code in an `else` statement. Previously, we didn't have this `else` since there wasn't the `if (options.chromecastMode)` conditional so it wasn't indented. Is that okay? I know it makes code review a little tougher but hopefully there's less diffs now :)

kdaswani:
Chromecast unit tests passing after latest commit:
Screen Shot 2021-06-07 at 5 21 01 PM

I am going to run through local testing this week, and then update this branch with master again before working on the process to pin a version to Fox's dev Chromecast source(s). Pooya helped me with a [different analytics.js PR](https://github.com/segmentio/analytics.js-integrations/pull/597) last week so I am familiar with analytics.js-versions and the pinning process now! Anything else I should take care of before we pin @mericsson?

kdaswani:
I started local testing and ran into issues with `Video Ad Started` not including an `AdBreakInfo` object and `Video Content Started` not including a `ChapterInfo` object. For example, Adobe debug logs show the following up `Video Content Started`:

```[2021-06-29T01:42:04.801Z] ADBMobile - Debug [Media] [MediaTracker (tracker1)] trackEvent(chapterStart info:undefined metadata:undefined)
```VM12:1 [2021-06-29T01:42:04.802Z] ADBMobile - Debug [Media] [MediaTracker (tracker1)] RuleEngine: Predicates failed for rule API::trackEvent(chapterStart) : ChapterInfo passed into 'API:trackEvent(ChapterStart)' is invalid.

We include these objects in the a.js integration, so we also need to include them in the Chromecast logic. I pinged @Isuru-F to discuss resolving this!

File: integrations/adobe-analytics/lib/index.js [hunk 21]
```diff
@@ -139,9 +147,90 @@ AdobeAnalytics.prototype.initialize = function() {
...
+        "timezone": "MDT",
+        "timezoneOffset": -360,
+        "referrerTimeout": 0,
+        "poi": []
+      },
+      // TO DO: Update placeholder values with settings or properties
+      "mediaHeartbeat": {
+        "server": `${options.heartbeatTrackingServerUrl}`,
+        "publisher": `${options.marketingCloudOrgId}`,
+        "channel": "test-channel-chromecast",
```
kdaswani, line 188: @Isuru-F any thoughts on if / how we could dynamically populate this? I think Fox only passes `channel` and `playerName` in the events, and I don't think those are available at this point. I also don't want to add a new setting in order to pipe something in. I can ask Fox if it matters what these values are for them, since the event metadata should indicate things like channel, player name, etc.? Do you have any indication on the point of these fields in the `mediaHeartbeat` config object?

kdaswani (commented):
Overall, this is looking good! Unit tests are passing and local testing checks out. I am going to do another round of local testing tomorrow or early next week. I added a comment for the last thing. And once we figure that out, I think we're ready to generate a version of a.js with this and pin it to Fox's FoxNews Chromecast source for testing. I can handle all of that!

Also, as for the automatically collected context fields not mapping, I think this is a non-issue at this point since Fox isn't mapping any of those on web anyways. In addition, Fox's Chromecast events don't appear to have anything they would want to map aside from `context.page` (which do get picked up) and `context.traits` (which they are manually adding on the client so those will get picked up).

integrations/adobe-analytics/lib/index.js
Isuru-F: @kdaswani  - from what I can tell this JSON object looks like it must exist in a script tag on init, I don't think its something that can be passed in the track calls. 

That said is channel/playername something that should be "hardcoded" to chromecast? Looking at the definition in the docs it says
["String that represents the name of the content distribution channel." / "String that represents the name of the player."](https://experienceleague.adobe.com/docs/media-analytics/using/sdk-implement/setup/set-up-chromecast.html?lang=en)

If it does need to be dynamically set we may need to revisit a few things and do some more testing on a few parts to see how it behaves without it on init. 

For reference this is a chromecast AA example app I look at for some of these questions https://github.com/Adobe-Marketing-Cloud/media-sdks/blob/master/sdks/chromecast/samples/BasicPlayerSample/index.html

kdaswani: Thanks @Isuru-F that was my sense too. Since it's in the initialization script, we won't have track calls yet to set values dynamically based on properties. I'm not super clear on what `channel` and `playerName` need to be (and how important they are) so I'm going to ask Fox for a copy of their existing Chromecast `ADBMobileConfig` since they integrate natively today. Hoping I can get that sorted with them next week and then we can consider this ready for testing. I'll keep you updated. Thanks for all your help!

kdaswani: Chatted with Fox. They don't report on these fields and are not concerned about falling back on "chromecast" or "unknown" for these since we do that in the web integration: https://github.com/segmentio/analytics.js-integrations/blob/72e95388ce84f56d1776a8e2e546eba984def318/integrations/adobe-analytics/lib/index.js#L1232. Hardcoding `channel` sdkVersion` and `playerName` as a result.

kdaswani:
I just spoke to our contact at Fox and she shared that this will need to be put on hold from the Fox side, as they do not have resources/bandwidth to test. Their goal is to test with FoxNews Chromecast first, but that team has not prioritized the work and she does not have a timeline for when they will. Once they are ready to test, we can generate an analytics.js version and pin this to their dev site. In the meantime, this is on hold and can sit as an open PR. Is that ok @mericsson?

mericsson:
> I just spoke to our contact at Fox and she shared that this will need to be put on hold from the Fox side, as they do not have resources/bandwidth to test. Their goal is to test with FoxNews Chromecast first, but that team has not prioritized the work and she does not have a timeline for when they will. Once they are ready to test, we can generate an analytics.js version and pin this to their dev site since this is ready for customer testing. In the meantime, this is on hold and can sit as an open PR. Is that ok @mericsson?

Sounds good. I updated it to a Draft PR to distinguish it as WIP.

kdaswani:
@mericsson change of plans, Fox is almost ready to test this! They created the source that we will pin the version of a.js with these changes to: https://app.segment.com/foxdcg/sources/fox_news_chromecast_development/overview. I will start the [process for pinning](https://paper.dropbox.com/doc/Releasing-Analytics.js--BOCVJQrZxwO63eHgkAjGNtKZAg-QYiKKIDvhsZk2c0bOTEhp) (i.e. analytics.js-private commit) early next week if you are comfortable with that. I think we can leave this PR as a draft since we don't plan to merge this into the existing integration at this point?

integrations/adobe-analytics/lib/chromecast-heartbeat.js
```diff
@@ -0,0 +1,202 @@
...
+
+var chromecastHeartbeat = {
+
+  /**
+  * Returns a mediaMetadata Object 
+  * @param {Track} track
+  * @returns {mediaMetadata} Maps the contextValue prop if it exists in the track.properties
+  */
+  extractMediaMetadata: function (track) {
+    debugger;
```
mericsson, line 11:
```suggestion
```

kdaswani: going to run more local test later today and then remove this!

kdaswani:
Pinned this to Fox's FoxNews Chromecast Dev and Prod sources on August 24, 2021.

PR #612, opened 2021-06-17
Add profitwell destination

pooyaj:
**What does this PR do?**
Adding the profitwell destination. Originally hosted here: https://github.com/ProfitWell/segment-dotjs but we need it internally to serve AJS 2.0. 

No breaking change

**Testing**
Testing not required as this is a carbon copy of the third party version.


pooyaj:
@nettofarah any gotchas in importing these destinations? I just copied the code directly.

nettofarah (approved):
LGTM!

PR #625, opened 2021-08-18
add isOuterScript to fullstory configs

jordanskole:
**What does this PR do?**

**Testing**

We have had to set this variable outside of segment for a while now, and it makes my devs lose trust in why we are using segment. You can read more about the config here: https://help.fullstory.com/hc/en-us/articles/360020622514-Can-FullStory-capture-content-that-is-presented-in-iframes-#h_01F1G33B40Q2TPQA8MA7SF8Y5P

Not applicable. 

Yes, it should mimic the debug setting only setting `window['_fs_is_outer_script'] = true;`

https://help.fullstory.com/hc/en-us/articles/360020622514-Can-FullStory-capture-content-that-is-presented-in-iframes-#h_01F1G33B40Q2TPQA8MA7SF8Y5P

### DIFF
--- a/integrations/fullstory/lib/index.js
+++ b/integrations/fullstory/lib/index.js
@@ -21,6 +21,7 @@ var FullStory = (module.exports = integration('FullStory')
   .option('trackNamedPages', false)
   .option('trackCategorizedPages', false)
   .option('trackPagesWithEvents', true)
+  .option('isOuterScript', false)
   .tag(
     '<script async src="https://edge.fullstory.com/s/fs.js" crossorigin="anonymous"></script>'
   ));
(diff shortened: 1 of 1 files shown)


lmakarov:
We need this feature as well. Looking forward for this to be merged!

Bogdan1988:
Our team is looking into similar functionality, we need ability to use FullStory inside an iframe>
Thanks

lmakarov:
This is what prevents us from using the native Segment FullStory integration. It's is tiny change that will make a huge difference for anyone trying to integrate FullStory via Segment in an iframed application.

cc @pooyaj

pooyaj (approved):

PR #626, opened 2021-08-24
Fix/moengage id reset handle

nawazMoEngage:
What does this PR do?
- it has bug fixes where reset was getting called when userID is passed
- alias method implemented

- No

- We were expecting the `reset()` method to be available in integration and wrote the code accordingly. Now its causing unnecessary logout from our SDK on adding user IDs.

- Not applicable

Does this require a new integration setting? If so, please explain how the new setting works - 
- No

Links to helpful docs and other external resources
- https://docs.moengage.com/docs/segment-websdk-integration

Why is this PR raised?
- There was an issue raised by a customer which was causing unnecessary logouts on MoEngage side and thus bloating the user counts in MoEngage

Are there any docs surrounding the change?
- https://docs.moengage.com/docs/segment-websdk-integration

What will the customer impact be as a result of the change?
- The customer experience will improve when integrating MoEngage and Segment as we will no longer create multiple users on login.

What will the customer impact be if the change is not implemented?
- The user count in MoEngage will bloat up when using Segment <> MoEngage integration and setting the user ID on identify.

Testing

### DIFF
--- a/integrations/moengage/lib/index.js
+++ b/integrations/moengage/lib/index.js
@@ -104,12 +104,6 @@ MoEngage.prototype.loaded = function() {
 
 MoEngage.prototype.identify = function(identify) {
   var self = this;
-  // Important: MoEngage require you to manually call reset to wipe the unique id for the session
-  // if you don't do this and call add_unique_user_id w/ a different userId, it will overwrite the previous user's
-  // unique id and all their traits so we need to manually check if it's a new user since `analytics.reset()` is not
-  // mapped for ajs integrations
-  // analytics.js regenerates anonymousId if you call `.identify()` with a unique userId value different from the cache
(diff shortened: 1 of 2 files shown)


RobbenWeemsSegment (approved):

RobbenWeemsSegment (approved):

coreyching:
@RobbenWeemsSegment Do you have an update on this PR? I see you approved changes. Checking in on the status

juliofarah (approved):

juliofarah:
HI @coreyching and @nawazMoEngage 
Thanks for putting this PR together!

As I needed to bump a minor version for this library I duplicated these changes into a new origin branch and have just merged the following: https://github.com/segmentio/analytics.js-integrations/pull/635 .

A new package should be available on NPM. We're going to start the roll out for customers using AJS Bundles as soon as possible.

Closing this.

PR #630, opened 2021-09-10
[STRATCONN-354] Marketo-V2 associateLead() deprecation

marinhero:
**What does this PR do?**

Deprecates the Marketo `associateLead` and uses the recommended method of `MktoForms2` instead.

Marketo will be removing  `associateLead` in October. If customers don't take preemptive actions to start using forms,  their calls to identify will be showing a message in the console: 

 'Invalid settings for identify method. Please review your Marketo V2 destination settings.'

Customer comms and doc updates are in progress to mitigate the hard cutover.  

**Testing**

- Testing completed successfully using the analytics.js-integrations compiler connected to stage.

Yes, two:

- marketoFormId: every form in Marketo has an ID. We need customers to provide their form ID so we can pull it and use the form in calls to identify.
- marketoHostUrl: the forms we will be using are relative to a domain. We need customers to provide their host URL so we can fetch the form from the right host.

Both settings are strings, available via CDN, and not required since track can work without them.

- https://developers.marketo.com/javascript-api/forms/
- https://developers.marketo.com/blog/deprecation-of-munchkin-associate-lead-method/
- https://developers.marketo.com/blog/make-a-marketo-form-submission-in-the-background/


integrations/marketo-v2/lib/index.js
```diff
@@ -98,17 +103,50 @@ var Marketo = (module.exports = integration('Marketo V2')
 
 Marketo.prototype.initialize = function() {
   var self = this;
```
brennan, line 105: I don't think you need to reassign `this` to `self` here. When passed as an argument to a method (as on line 108), `this` should still refer to the outer context (in this case, the `initialize` prototype method).

integrations/marketo-v2/lib/index.js
```diff
@@ -82,11 +82,16 @@ var apiNameMapping = {
 var Marketo = (module.exports = integration('Marketo V2')
   .assumesPageview()
   .global('Munchkin')
+  .global('MktForms2')
   .option('host', 'https://api.segment.io')
   .option('accountId', '')
   .option('projectId', '')
   .option('traits', [])
```
brennan, line 89: Do we need to add the `marketoFormId` option somewhere here?

integrations/marketo-v2/lib/index.js
```diff
@@ -4,13 +4,13 @@
...
  */
 
-var fmt = require('@segment/fmt');
 var integration = require('@segment/analytics.js-integration');
 var is = require('is');
-var jsonp = require('jsonp');
 var url = require('component-url');
 var when = require('do-when');
 var each = require('@ndhoule/each');
+var onBody = require('on-body-ready');
```
brennan, line 12: I wonder if we simply write our own helper method rather than install a third-party package. The relevant code looks to be about ~10 lines or so. Can we copy/paste?

marinhero: Good idea! I was just following a pattern I saw in another integration but I'd rather not use a dependency. Thanks!

integrations/marketo-v2/lib/index.js
marinhero: Good call, I'm not sure but I will learn about this. There's also `marketoHostURL` and it should be added in here as well

integrations/marketo-v2/lib/index.js
marinhero: Turns out `load` will be enough to make sure the form library is ready to work! we don't need to bring anything in 😎

integrations/marketo-v2/lib/index.js
marinhero: Wouldn't that only apply if you are trying to access `this` itself inside of the function? All I'm doing on 108 is to pass a reference to the value inside the options object. Options become available via the scope of the function (`this`).  

I'll move the reasign below the variable definitions but please let me know what you think.

integrations/marketo-v2/lib/index.js
```diff
@@ -141,6 +177,16 @@ Marketo.prototype.page = function(page) {
   });
 };
 
+Marketo.prototype.setupAndSubmitForm = function(traits, form) {
+  console.log('PRIME')
```
lcampos, line 181: Do we need to keep this console.log ?

integrations/marketo-v2/lib/index.js
```diff
@@ -97,18 +103,48 @@ var Marketo = (module.exports = integration('Marketo V2')
...
     // marketo integration actually loads a marketo snippet
     // and the snippet loads the real marketo, this is required
     // because there's a race between `window.mktoMunchkinFunction = sinon.spy()`
     // and marketo's real javascript which overrides `window.mktoMunchkinFunction`
     // and deletes the spy.
     when(self.loaded, self.ready);
   });
+
+  this.load('forms', { marketoHostUrl: marketoHostUrl }, function() {
+    var marketoForm = document.createElement('form');
```
lcampos, line 138: Should we only create the form id `marketoHostUrl` is defined ?

compiler/package.json
```diff
@@ -70,7 +70,6 @@
     "@segment/analytics.js-integration-gtag": "file:../integrations/gtag",
     "@segment/analytics.js-integration-gosquared": "file:../integrations/gosquared",
     "@segment/analytics.js-integration-heap": "file:../integrations/heap",
-    "@segment/analytics.js-integration-hellobar": "file:../integrations/hellobar",
```
lcampos, line 73: These deletes seem unrelated to your changes, do we need them in this PR ?

integrations/marketo-v2/lib/index.js
marinhero: No no, this was for testing purposes. My bad!

integrations/marketo-v2/lib/index.js
marinhero: There's a validation for that already. The code won't reach this point if `marketoHostUrl` is undefined

integrations/marketo-v2/lib/index.js
```diff
@@ -97,18 +103,48 @@ var Marketo = (module.exports = integration('Marketo V2')
  */
 
 Marketo.prototype.initialize = function() {
+  var munchkinId = this.options.accountId;
+  var marketoHostUrl = this.options.marketoHostUrl;
+  var marketoFormId = parseInt(this.options.marketoFormId, 10);
+
+  var identifySettingsAreInvalid =
```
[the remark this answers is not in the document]
marinhero: Good idea! I'll take that into consideration because Marketo provides the URL without protocol.

compiler/package.json
marinhero: Indeed it's cleanup work but I think we should keep them. Otherwise, people will always have to troubleshoot the compiler before being able to start working.

integrations/marketo-v2/HISTORY.md
```diff
@@ -1,3 +1,11 @@
+4.0.0 / 2021-09-09
```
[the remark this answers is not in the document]
marinhero: Good catch. I've just added it.

integrations/marketo-v2/lib/index.js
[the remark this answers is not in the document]
marinhero: Now that I took a closer look, I think we should address this with a validation for the setting (DB level) and not as part of the integration. I think I'll leave it as is. What do you think?

brennan:
@marinhero Is it possible to share, either directly in the PR comment or in a linked paper doc, the manual testing you did using the compiler tool?

integrations/marketo-v2/lib/index.js
```diff
@@ -141,6 +177,15 @@ Marketo.prototype.page = function(page) {
   });
 };
 
+Marketo.prototype.setupAndSubmitForm = function(traits, form) {
+  form.addHiddenFields(traits, form);
+  // Do not remove this callback. This ensures there are no page refreshes after the form is submitted.
+  form.onSuccess(function() {
```
marinhero, line 183: Add an onFailure callback to not refresh in that scenario either

lcampos (approved):
Approved pending the couple of changes we talked about earlier

integrations/marketo-v2/lib/index.js
marinhero: No callback available https://developers.marketo.com/javascript-api/forms/api-reference/ :|

PR #650, opened 2022-01-09
support Totango EU services

shmulikgolan:
**Note: This change needed to be supported with a new selection option, region, with the values "us" or "eu1"**

**What does this PR do?**
adding region field to Totango init. 

no

**Testing**

### DIFF
--- a/integrations/totango/lib/index.js
+++ b/integrations/totango/lib/index.js
@@ -44,10 +44,14 @@ Totango.prototype.initialize = function(page) {
     track:function(t,o,n,a){window.totango_tmp_stack.push({activity:t,module:o,org:n,user:a}); return -1;}};
   /* eslint-enable */
 
+  /* Int Totango JS */
+  var defaultRegion = 'us'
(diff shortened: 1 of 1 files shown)


integrations/totango/lib/index.js
```diff
@@ -44,10 +44,14 @@ Totango.prototype.initialize = function(page) {
     track:function(t,o,n,a){window.totango_tmp_stack.push({activity:t,module:o,org:n,user:a}); return -1;}};
   /* eslint-enable */
 
+  /* Int Totango JS */
+  var defaultRegion = 'us'
+  var selected_region = this.options.region ?  this.options.region : defaultRegion;
```
pooyaj, line 49: @shmulikgolan please define the `region` option on the code block of line 17-25. You can also default the option to `us` on the same code block rather than doing it on line 48.

pooyaj (changes requested):
Thanks for the PR. Added a minor suggestion.

shmulikgolan:
Created new PR (652) please delete this one.

PR #657, opened 2022-02-09
[STRATCONN-1076] Add deprecation notice to Adobe Target

marinhero:
**What does this PR do?**

Add a deprecation warning to the Adobe Target integration.

No, no customers are using this integration.

**Testing**

- Testing not required because this only adds comments into the codebase.

StratConn is building a new target destination in actions, that's why we are marking this as deprecated.


kdaswani:
looks great, thank you!

kdaswani (approved):

PR #662, opened 2022-02-18
Add Woopra auto-track configurations and misc fixes

eliekhoury:
**What does this PR do?**
This PR adds support for the new `woopra.config` option `click_tracking` and removes the outdated `ping` configuration which is no longer supported.

This PR also removes having to stringify array properties as it's now automatically handled by the Woopra Javascript SDK.

No

**Testing**
Testing completed successfully

```
    ✓ should have the right settings
    before loading
      #initialize
        ✓ should create a woopra object
        ✓ should configure woopra
        ✓ should not send options if they are null, or empty
        ✓ should call #load
    loading
      ✓ should load
    after loading
      #page
        ✓ should send a page view
        ✓ should send a title
        ✓ should prefer a name
        ✓ should prefer a category and name
        ✓ context is set
      #identify
        ✓ should send an id
        ✓ should send traits
        ✓ should send an id and traits
        ✓ should alias the name properly
        ✓ should should convert trait dates to unix timestamp in milliseconds
        ✓ context is set
      #track
        ✓ should send an event
        ✓ should send properties
        ✓ context is set

HeadlessChrome 98.0.4758 (Mac OS X 10.15.7): Executed 20 of 20 SUCCESS (0.265 secs / 0.049 secs)
TOTAL: 20 SUCCESS
```

No

This update requires adding a new checkbox to the Woopra configuration in the Segment portal to enable/disable `click_tracking`. 

https://docs.woopra.com/reference/woopraconfig#auto-tracking-configs


integrations/woopra/lib/index.js
```diff
@@ -22,14 +22,11 @@ var Woopra = (module.exports = integration('Woopra')
...
-  .option('pingInterval', 12000)
   .option('idleTimeout', 300000)
-  .option('downloadTracking', true)
-  .option('outgoingTracking', true)
+  .option('downloadTracking', false)
+  .option('outgoingTracking', false)
+  .option('clickTracking', false)
   .option('outgoingIgnoreSubdomain', true)
-  .option('downloadPause', 200)
-  .option('outgoingPause', 400)
```
chrisradek, line 32: I see both of these in the GitHub readme configuration table - have they actually been removed from the library? It looks like the integration's default for `downloadPause` matches woopra's, but `outgoingPause` is double what woopra uses by default, so unless they've been removed and woopra no longer reads these settings we should consider leaving these in place.

integrations/woopra/lib/index.js
```diff
@@ -22,14 +22,11 @@ var Woopra = (module.exports = integration('Woopra')
   .option('cookieName', 'wooTracker')
   .option('cookieDomain', null)
   .option('cookiePath', '/')
-  .option('ping', true)
-  .option('pingInterval', 12000)
   .option('idleTimeout', 300000)
-  .option('downloadTracking', true)
-  .option('outgoingTracking', true)
```
chrisradek, line 29: Changing the default behavior for `downloadTracking` and `outgoingTracking` would be a breaking change for existing users of this integration. Can these be reverted back to `true` by default or does it cause issues with `clickTracking`?

integrations/woopra/lib/index.js
```diff
@@ -22,14 +22,11 @@ var Woopra = (module.exports = integration('Woopra')
   .option('cookieName', 'wooTracker')
   .option('cookieDomain', null)
   .option('cookiePath', '/')
-  .option('ping', true)
-  .option('pingInterval', 12000)
```
chrisradek, line 26: I see the ping options are deprecated by the woopra library - just curious, would a user interact with the woopra library any differently if these are turned on?

integrations/woopra/HISTORY.md
```diff
@@ -1,4 +1,12 @@
 
+
+2.1.0 / 2022-02-18
```
chrisradek, line 3: I think since we're no longer passing through some options to `woopra` this should be considered a major update.

integrations/woopra/lib/index.js
```diff
@@ -139,35 +147,9 @@ Woopra.prototype.identify = function(identify) {
 
 Woopra.prototype.track = function(track) {
   setContext(track);
-  window.woopra.track(track.event(), stringifyNested(track.properties()));
```
chrisradek, line 142: Awesome that we don't have to stringify this anymore!

chrisradek (commented):
@eliekhoury Thank you for creating this PR! Forgive all my questions - I'm not too familiar with `woopra` yet! I want to understand if the options that are removed were also removed or deprecated in `woopra`, and if we _need_ to change the default behavior for the options that weren't removed.

integrations/woopra/lib/index.js
eliekhoury: Those have been removed. We're now using beacons to track downloads and outgoing links. The pause is no longer needed.

integrations/woopra/lib/index.js
eliekhoury: The default configuration of the JS tracker is false for both these features. When it was true, Segment was assuming that those features are already enabled by default and was not including them in the injected javascript configuration. So switching the default to false fixes this issue (if my understanding is correct).

integrations/woopra/lib/index.js
eliekhoury: The ping mechanism in Woopra is no longer supported. It has been replaced with a more robust mechanism using Beacons. So the ping and pingInterval configurations are ignored right now by the Woopra Javascript SDK. So, they should be removed from Segment's configuration. 

Users will not need to interact with the library any differently.

integrations/woopra/HISTORY.md
eliekhoury: That's up to Segment to decide. We're happy to bump this to 3.0.

eliekhoury:
@chrisradek Thanks Chris for the thorough review. While we fixed the default settings in the Woopra tracker, the default Javascript snippet we generate is now enabling the three Auto Track features by default (`click_tracking`, `download_tracking` and `outgoing_tracking`). See screenshot:

Screen Shot 2022-02-23 at 12 46 33 PM

**It would be ideal if Segment can turn those on for all new installations while still allowing users to opt-out from any of them. Partner support mentioned that this can be done on the Portal side.**

integrations/woopra/lib/index.js
```diff
@@ -139,35 +147,9 @@ Woopra.prototype.identify = function(identify) {
...
- * Undocumented aspect of Woopra's API, but apparently required. Breaks
- * on `Completed Order` `properties.products`.
- *
- * @param {Object} obj
- * @return {Object}
- * @api private
- */
-
-function stringifyNested(obj) {
-  return foldl(
```
chrisradek, line 157: Can you also remove the `@ndhoule/foldl` import now that we don't need to use it?

integrations/woopra/lib/index.js
chrisradek: We're actually calling `config` on each option when we initialize the integration:
https://github.com/segmentio/analytics.js-integrations/blob/61bd44991705cf4f07937a626cb70cb8aec2404c/integrations/woopra/lib/index.js#L77-L82
So the current behavior would be that these are both set to `true` by default.

integrations/woopra/lib/index.js
eliekhoury: Sure

integrations/woopra/lib/index.js
eliekhoury: Thanks for the clarification. I was under the impression that Segment is only sending a config update for properties that do not match the default setting. I'll switch those to true.

integrations/woopra/test/index.test.js
```diff
@@ -38,14 +38,11 @@ describe('Woopra', function() {
         .option('cookieName', 'wooTracker')
         .option('cookieDomain', null)
         .option('cookiePath', '/')
-        .option('ping', true)
-        .option('pingInterval', 12000)
         .option('idleTimeout', 300000)
         .option('downloadTracking', true)
         .option('outgoingTracking', true)
+        .option('clickTracking', true)
```
chrisradek, line 44: I don't actually see what the default value for `clickTracking` is on the woopra docs. If it's already `true` then this is totally fine.

chrisradek (commented):
This looks great, thank you! One question about the default value for `clickTracking`. Since this will be rolled out to existing users as well we should have the default for the new option match whatever is the default for woopra. This will still be overridden by whatever the user selects when configuring woopra in the segment app.

integrations/woopra/test/index.test.js
eliekhoury: Good point. `clickTracking` is not turned on by default in the tracker for retroactive reasons. But we would like to be turned on for every new installation. Is it possible to turn it off for existing installations and automatically turn it on for new installations?

chrisradek: Sorry for the delayed response - it looks like it isn't possible to turn it off for existing installs and have it defaulted on for new installations. We would need to set it as `false` by default so that it doesn't change existing behavior.

chrisradek (approved):
@eliekhoury Changes look great! I'm merging into a feature branch so I can trigger CI, then once that runs will merge into master.

Thank you so much for contributing and working with me on this!

PR #665, opened 2022-03-02
Encoded integration names

arielsilvestri:
This supports building and uploading obfuscated bundles to our production S3 bucket for destinations. 

Sourcemaps will not be obfuscated as a result of these changes, so that they can still be found and loaded by the bundles downloaded by browsers.

Testing completed successfully with uploads and deployments via the staging branch, and successfully loaded in the snippet example in `analytics-next`, associated PR to come once this is merged and deployed.

## Examples
S3 directory for kissmetrics: 
Screen Shot 2022-03-02 at 3 59 52 PM

S3 directory for obfuscated kissmetrics:
Screen Shot 2022-03-02 at 4 01 26 PM


scripts/upload-assets.js
```diff
@@ -122,6 +125,19 @@ async function uploadAssets() {
       true
     );
 
+    await putObject(
+      fileName,
+      `${type}/${bufferClientName}/${version}/${file.includes('map') ? file : bufferFile}`,
```
chrisradek, line 130: It might be safer to add a `.` prefix to `map` just in case we ever add an integration that has `map` in its name (we don't currently, but you never know!)
```suggestion
      `${type}/${bufferClientName}/${version}/${file.includes('.map') ? file : bufferFile}`,
```

scripts/upload-assets.js
```diff
@@ -107,9 +106,13 @@ async function uploadAssets() {
     }
 
     const f = fileName.split('/');
+
     const type = f[1]; // get integration name
     const integration = f[2]; // get integration name
+    const clientName = f[f.length - 1].substring(0, f[f.length-1].indexOf('.'));
+    const bufferClientName = Buffer.from(clientName).toString('base64').replace(/=/g, '');
```
chrisradek, line 113: If you move this below `const file = f[f.length - 1]`, you can simplify to:
```ts
    const clientName = file.substring(0, file.indexOf('.'));
```

chrisradek (commented):
Really like your idea of base64 encoding the integration name. Changes look good, just had some minor comments.

scripts/upload-assets.js
```diff
@@ -122,6 +125,19 @@ async function uploadAssets() {
...
 
+    await putObject(
+      fileName,
+      `${type}/${bufferClientName}/${version}/${file.includes('.map') ? file : bufferFile}`,
+      true
+    );
+
+    await putObject(
+      fileName,
+      `${type}/${bufferClientName}/latest/${file.includes('map') ? file : bufferFile}`,
```
chrisradek, line 136: I didn't point out this location but can you update this to `.map` as well?

pooyaj (approved):
🙌

chrisradek (approved):
Looks great, :ship::it:

PR #692, opened 2022-07-26
add support for snake_case for contentAssetId in Adobe Heartbeat

rhall-twilio:
**What does this PR do?**
* Drop support for `assetId` for `StandardVideoMetadata` in Adobe Heartbeat
* Add support for `content_asset_id` for `StandardVideoMetadata` in Adobe Heartbeat

The removal of `assetId` from `StandardVideoMetadata` cannot be considered a breaking change given that the value was always overriden by `content_asset_id`, even if undefined. See for more details:  https://segment.atlassian.net/browse/STRATCONN-1491

**Testing**

Testing completed successfully using the `tester` page by stepping through debugger on the `createStandardVideoMetadata` function and making sure that

 ```analytics.track("video playback started", {assetId: "content-asset-id"})```

did not populate the media asset id while

 ```analytics.track("video playback started", {content-asset-id: "content-asset-id"})```

did successfully so

Testing completed successfully using unit tests.

https://segment.atlassian.net/browse/STRATCONN-1491

N/A

N/A

https://segment.com/docs/connections/destinations/catalog/adobe-analytics/heartbeat/#video-playback-started

### DIFF
--- a/integrations/adobe-analytics/lib/index.js
+++ b/integrations/adobe-analytics/lib/index.js
@@ -1501,7 +1501,7 @@ function createStandardVideoMetadata(track, mediaObj) {
   adbSegMap[metaKeys.SHOW] = ["program"],
   adbSegMap[metaKeys.SEASON] = ["season"],
   adbSegMap[metaKeys.EPISODE] = ["episode"],
-  adbSegMap[metaKeys.ASSET_ID] = [ "content_asset_id", "contentAssetId"],
+  adbSegMap[metaKeys.ASSET_ID] = ["contentAssetId", "content_asset_id"],
   adbSegMap[metaKeys.GENRE] = ["genre"],
   adbSegMap[metaKeys.FIRST_AIR_DATE] = ["airdate"],
   adbSegMap[metaKeys.ORIGINATOR] = ["publisher"],
--- a/integrations/adobe-analytics/lib/index.js
+++ b/integrations/adobe-analytics/lib/index.js
@@ -1497,23 +1497,30 @@ function createStandardVideoMetadata(track, mediaObj) {
   var props = track.properties();
   var metaKeys = videoAnalytics.MediaHeartbeat.VideoMetadataKeys;
   var stdVidMeta = {};
-  var segAdbMap = {
-    program: metaKeys.SHOW,
-    season: metaKeys.SEASON,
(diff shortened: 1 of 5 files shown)


integrations/adobe-analytics/lib/index.js
```diff
@@ -1497,23 +1497,30 @@ function createStandardVideoMetadata(track, mediaObj) {
...
-    airdate: metaKeys.FIRST_AIR_DATE,
-    publisher: metaKeys.ORIGINATOR,
-    channel: metaKeys.NETWORK,
-    rating: metaKeys.RATING
-  };
+  var adbSegMap = {}
+  adbSegMap[metaKeys.SHOW] = ["program"],
+  adbSegMap[metaKeys.SEASON] = ["season"],
+  adbSegMap[metaKeys.EPISODE] = ["episode"],
+  adbSegMap[metaKeys.ASSET_ID] = [ "content_asset_id", "contentAssetId"],
```
kdaswani, line 1504: Do this logic prioritize `content_asset_id` or `contentAssetId` first? I think we want to always prioritize `contentAssetId` as to not break any existing implementation (i.e. if the customer happens to have both snake and camel case in their event - unlikely but good to be safe). Fallback on `content_asset_id`.

rhall-twilio: It is currently prioritizing `content_asset_id`. I will make the correction.

brennan (approved):

integrations/adobe-analytics/lib/index.js
kdaswani: The test below makes me think that maybe these got swapped. Can we swap them back @rhall-twilio? Otherwise all looks good to me! Will wait for local testing to be complete and review from another eng before approving. Thanks!

kdaswani (changes requested):

integrations/adobe-analytics/lib/index.js
kdaswani: So fast at responding @rhall-twilio 😂

rhall-twilio: This is now resolved.

kdaswani (approved):
ty robert! looks great

PR #697, opened 2022-09-06
Adding regional api support for intercom

pooyaj:
**What does this PR do?**
Adding an `apiBase` option for intercom. Defaulting to empty, but if user chose to pick an API region, then we will respect that value when initializing intercom. 

No breaking change

**Testing**
Testing completed successfully on local using local integrations tester

Intercom is adding regional support for data processing, so we are providing an option to pick region for Intercom customers. 

NA

Yes, we need to add `apiBase` to partner portal. It will be a choice between the following values: 
```
      choices: [
        {
          label: 'US',
          value: 'https://api-iam.intercom.io'
        },
        {
          label: 'EU',
          value: 'https://api-iam.eu.intercom.io'
        },
        {
          label: 'Australia',
          value: 'https://api-iam.au.intercom.io'
        }
      ],
```

NA


integrations/intercom/lib/index.js
```diff
@@ -23,6 +23,7 @@ var Intercom = (module.exports = integration('Intercom')
   .option('activator', '#IntercomDefaultWidget')
   .option('appId', '')
   .option('richLinkProperties', [])
+  .option('apiBase', '')
```
chrisradek, line 26: So Intercom isn't planning to make this required right? Wondering if we should default to the current value (US?) - what does the option look like in the app if you have 3 choices but none selected?

chrisradek (approved):
Looks great! Had 1 question but it's not a blocker either way.

integrations/intercom/lib/index.js
pooyaj: Actually originally added the default (`https://api-iam.intercom.io`) here, but then thought that's not 100% backwards compatible, because it forces users to use `https://api-iam.intercom.io`, while the old behavior doesn't specify the API host explicitly. Although I'm not 100% sure if defaulting to `https://api-iam.intercom.io` would cause any issues, but if somehow some users override the API host, and then we override it again it will be bad.

chrisradek: Gotcha, makes sense!

PR #699, opened 2022-09-09
update all @segment/trample to same minimum version

chrisradek:
This PR updates the `@segment/trample` dependency to have the same minimum version across all integrations (currently just 2 - adobe analytics and kenshoo infinity).

This is needed because the common vendor file that contains the dependencies of all integrations was only pulling in the older version of `@segment/trample`. I'm not sure why it chose `0.2.0` instead of `0.2.1` since the latter met the version requirements for kenshoo, but this change resolves the issue.

This is especially important for the adobe analytics integration which needs the `0.2.1` update.

**Testing**
Testing completed successfully by rebuilding the integrations and validating the new trample code exists in the vendor file.


pooyaj (approved):
weird issue ... nice find 💥

Issue #703, opened 2022-11-08
Braze deduplication middleware not available when using obfuscated bundle fetches


cysp:
Analytics.js 2.0 fails to fetch the Braze deduplication middleware (`analytics.js-middleware-braze-deduplicate`) when obfuscation is enabled.
This appears to be caused by a bug in the asset uploading script in this repository which assumes that there are no interior `.` characters in the base part of integration filenames:
https://github.com/segmentio/analytics.js-integrations/blob/f8cc4442c9ed975d8bb317e78ddec277014baf6e/scripts/upload-assets.js#L113-L115

Analytics.js 2.0 expects to find the integration script at `https://cdn.segment.com/next-integrations/middleware/YW5hbHl0aWNzLmpzLW1pZGRsZXdhcmUtYnJhemUtZGVkdXBsaWNhdGU/latest/YW5hbHl0aWNzLmpzLW1pZGRsZXdhcmUtYnJhemUtZGVkdXBsaWNhdGU.js.gz` (the obfuscated equivalent of `https://cdn.segment.com/next-integrations/middleware/analytics.js-middleware-braze-deduplicate/latest/analytics.js-middleware-braze-deduplicate.js.gz`), which currently gives a 404 response.

The obfuscation logic in the uploading script in this repository has resulted in the file incorrectly being uploaded as `https://cdn.segment.com/next-integrations/middleware/YW5hbHl0aWNz/latest/YW5hbHl0aWNz.js-middleware-braze-deduplicate.js.gz`

PR #709, opened 2022-11-17
Check for last instance of .js extension in filename

arielsilvestri:
**What does this PR do?**
- Accounts for the existence of `.js` in the filename of a given integration, per https://segment.atlassian.net/browse/LIBWEB-1278

**Testing**

To prevent having to change the structure of a number of filenames/potentially breaking existing functionality, I scoped this change to account for a middleware that's prefix'ed with `analytics.js` in its filename.


cysp:
Sweet, looks like this will fix #703

scripts/upload-assets.js
```diff
@@ -110,7 +110,9 @@ async function uploadAssets() {
     const type = f[1]; // get integration name
     const integration = f[2]; // get integration name
     const file = f[f.length - 1];
-    const clientName = file.substring(0, file.indexOf('.'));
+    const clientName = file.includes('analytics.js-') ? 
+      file.substring(0, file.lastIndexOf('.js')) :
+      file.substring(0, file.indexOf('.'));
```
silesky, line 115: Is there any reason to not just do this -- basically, we grab anything before the last .js?

```
const clientName  = file.substring(0, file.lastIndexOf('.js'))
```

arielsilvestri: That would affect the way our `.dynamic.js.gz` files are obfuscated and uploaded, which would require a change in how we fetch those files in ajs. It's doable, just would require a bit of a larger change to support and would still have some extra logic, just elsewhere.

silesky: ah K... I didn't realize that was a case we had to handle. Be nice to add a comment in the code about what the expected filename formats are for future reference, since there are no tests.

arielsilvestri: Done @silesky !

scripts/upload-assets.js
```diff
@@ -110,7 +110,11 @@ async function uploadAssets() {
     const type = f[1]; // get integration name
     const integration = f[2]; // get integration name
     const file = f[f.length - 1];
-    const clientName = file.substring(0, file.indexOf('.'));
+    // Supporting filenames where analytics.js is prepended to the bundle name.
+    // Filename formats typically end in .js or .js.gz
```
silesky, line 114:
```suggestion
    /**
     * Supporting filenames where analytics.js is prepended to the bundle name.
     * Filename formats typically end in .js or .js.gz
     * For example:
     * - asfghemUtZGVkdXBsaWNhdGU.js.gz
     * - analytics.js-middleware-braze-foo.js
     * - analytics.js-middleware-braze-foo.js.gz
     * - analytics.js-middleware-braze-foo.dynamic.js.gz
     */
```

silesky (approved):

PR #720, opened 2022-12-14
Bug/sdk 2455/fix sdk for turkish servers

sinhaayush7:
**What does this PR do?**
- This PR updates the clevertap sdk url

- No

**Testing**
- Testing not required as this PR updates the URL, and points to the latest build of clevertap web sdk.

- The existing clevertap sdk url pointed out to a deprecated version. This PR fixes this.

- NA

- No


nick-Ag (approved):

integrations/clevertap/lib/index.js
```diff
@@ -17,11 +17,8 @@ var CleverTap = (module.exports = integration('CleverTap')
   .global('clevertap')
   .option('clevertap_account_id', '')
   .option('region', '')
-  .tag('http', '<script src="http://static.clevertap.com/js/a.js">')
-  .tag(
```
pooyaj, line 21: There is a logic further down the code (in `initialize` method) that does this: 
```
  var protocol = useHttps() ? 'https' : 'http';
  this.load(protocol, this.ready);
```
the `this.load` method uses `protocol` to decide which script to load, but with this change we removed the `https` script, and therefore, the `this.load` will fail for `https` cases.

pooyaj (changes requested):

integrations/clevertap/lib/index.js
sinhaayush7: hi @pooyaj , If i understood this right, then the `this.load` will fail for http, since in the tag I've changed it to `https` now. The intent is to load the script over `https` only, since most of the people use https.

Also I think `use-https` library no longer serves the purpose, so would be removing this due to above stated reasons. 

Thanks!

pooyaj (approved):
Looks great!

PR #729, opened 2023-01-11
Adds segment track support in hellobar client script.

ayushgupta29:
**What does this PR do?**
This PR updates the hellobar script integration and provides the customer ability to use segment track events in hellobar event triggers.

No.

**Testing**
- Testing completed successfully using dev environment. A popup is tied to an event named `segment_event_1` which can be triggered successfully using `analytics.track("segment_event_1");` which validates the use case here.

We have hellobar popups being triggered on custom events. The customers who already use segment would now be able to use segment track events without integrating hellobar events directly by hooking into the segment events.

NA

No.

https://hellobarassist.freshdesk.com/support/solutions/articles/44002393650-triggering-popups-and-bars-on-a-user-event

### DIFF
--- a/integrations/hello-bar/HISTORY.md
+++ b/integrations/hello-bar/HISTORY.md
@@ -1,3 +1,8 @@
+3.0.2 / 2023-01-10
+==================
+
+  * Hook segment track events to hellobar event trigger
+
 3.0.0 / 2017-06-06
 ==================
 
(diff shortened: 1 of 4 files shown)


ayushgupta29:
Hey @pooyaj @nick-Ag - Not sure how to start the review process. Can you please help with this.

PR #734, opened 2023-02-14
Allow configuring heap integration settings

mwudka:
**What does this PR do?**
The current heap integration has two limitations which are proving challenging for our segment deployment:
* the snippet is loaded from a hardcoded URL, so we can't configure it to load the heap JS from our CDN
* there is no way to specify any heap configuration options, so we can't control the options like `secureCookie` or `disableTextCapture`.

This PR adds additional configuration options to the heap integration to support changing both.

No. 

**Testing**
I used the (very slick and excellent) local runner to test manually. I didn't see an easy way to add any tests, since it's hard to mock the underlying heap object. I'm also not sure the tests would be hugely valuable, since they would essentially be asserting that the code exists as written rather than really testing behavior. I am open to suggestions about a better way to handle this!

No

N/A

This change introduces two new optional settings:
1. `hostname` (`string`): Allows changing the URL from which the heap snippet is loaded. It defaults to "cdn.heapanalytics.com", the current hard-coded value, to preserve backwards compatibility. 
2. `options` (`object`): Allows passing an arbitrary set of configuration options to heap. They are passed to the heap JS code as is. Defaults to `{}` to preserve backwards compatibility.

[Heap JS options](https://developers.heap.io/docs/web#advanced-configurations)

### DIFF
--- a/integrations/heap/lib/index.js
+++ b/integrations/heap/lib/index.js
@@ -21,7 +21,9 @@ var toString = Object.prototype.toString; // in case this method has been overri
 var Heap = (module.exports = integration('Heap')
   .global('heap')
   .option('appId', '')
-  .tag('<script src="//cdn.heapanalytics.com/js/heap-{{ appId }}.js">'));
+  .option('hostname', 'cdn.heapanalytics.com')
+  .option('options', {})
+  .tag('<script src="//{{ hostname }}/js/heap-{{ appId }}.js">'));
 
 /**
  * Initialize.
@@ -61,7 +63,7 @@ Heap.prototype.initialize = function() {
     });
   };
 
-  window.heap.load(this.options.appId);
+  window.heap.load(this.options.appId, this.options.options);
   this.load(this.ready);
 };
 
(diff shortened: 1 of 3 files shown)


integrations/heap/lib/index.js
```diff
@@ -21,7 +21,9 @@ var toString = Object.prototype.toString; // in case this method has been overri
 var Heap = (module.exports = integration('Heap')
   .global('heap')
   .option('appId', '')
-  .tag('<script src="//cdn.heapanalytics.com/js/heap-{{ appId }}.js">'));
+  .option('hostname', 'cdn.heapanalytics.com')
```
mwudka, line 24: There seems to be some variation in how different integrations name this sort of parameter, so I wasn't sure of the best name. I'm happy to change to something else.

integrations/heap/lib/index.js
```diff
@@ -21,7 +21,9 @@ var toString = Object.prototype.toString; // in case this method has been overri
...
+  .option('options', {})
```
mwudka, line 25: The heap JS docs list some of the options, but there are additional options which are not documented. That leads me to believe that it might be best to just provide a single options object, since the allowed options might be account-specific and it might be confusing to provide some configuration options here that aren't available to all accounts. However, I'm happy to split this out into specific configuration options.

Issue #282, opened 2019-11-26
[google-tag-manager] Pull Request #15: Gracefully fail on window.dataLayer name conflict


SegmentDestinationsBot:
_Migrated from [#15](https://github.com/segment-integrations/analytics.js-integration-google-tag-manager/pull/15) by @Peripheral1994_

_I'm updating the body of this PR to do some testing, don't mind me_ - Carlos

PR #758, opened 2023-05-18
Added queueName option for custom dataLayer

ccook10-chwy:
**What does this PR do?**
Google Tag Manager supports custom queue (dataLayer) name but the existing integration does not allow for this. This change adds a `queueName` option to the GTM integration to support this.

No. This change adds an optional option which defaults to `dataLayer` so that it will continue to function as prior to the change.

**Testing**
Testing completed successfully. Unit tests were updated to test the new queueName option.

This change is needed to aid migration from on-site GTM to Segment destination for large or complex websites which cannot implement the migration all at once. Without this change, a GTM destination cannot be added to a Segment source without breaking an existing GTM implementation.

No

Yes.
- DataLayer
- You can specify a custom name for your dataLayer variable if you need to (e.g., coexist with an existing GTM implementation). The name must contain only letters, numbers, underscore, or dollar sign to meet JavaScript identifier rules.

### DIFF
--- a/integrations/google-tag-manager/HISTORY.md
+++ b/integrations/google-tag-manager/HISTORY.md
@@ -1,3 +1,7 @@
+2.6.0 / 2023-05-17
+==================
+  * Add support for queueName (custom dataLayer) option
+
 2.5.0 / 2017-04-27
 ==================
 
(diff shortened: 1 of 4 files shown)


ccook10-chwy:
@CarlosMecha please review, or advise what the process is for requesting a review. This change is partly inspired by this issue.

https://github.com/segmentio/analytics.js-integrations/issues/282

PR #759, opened 2023-06-01
Add option to choose a custom bucket

paco-walkme:
New option for certain customers to have a private bucket inside AWS

Testing completed successfully via unit tests.

### DIFF
--- a/integrations/walkme/lib/index.js
+++ b/integrations/walkme/lib/index.js
@@ -19,7 +19,7 @@ var WalkMe = module.exports = integration('WalkMe')
   .option('trackWalkMeEvents', false)
   .option('loadWalkMeInIframe', false)
   .option('integrityHash', '')
-  .option('customDirectory', 'users')
+  .option('customDirecotry', 'users')
   .tag('<script async="true" src="{{ url }}" crossorigin="" integrity="{{ hash }}">')
 
 /**
@@ -68,7 +68,7 @@ WalkMe.prototype.initialize = function() {
     sriSuffix = 'private_';
   }
 
-  var bucket = (this.options.customDirectory) ? this.options.customDirectory : 'users';
+  var bucket = (this.options.customDirecotry) ? this.options.customDirecotry : 'users';
   var url = 'https://cdn.walkme.com/' + bucket + '/' + walkMeSystemId + '/' + env + '/walkme_' + sriSuffix + walkMeSystemId + '_https.js';
 
   this.load({
--- a/integrations/walkme/lib/index.js
+++ b/integrations/walkme/lib/index.js
@@ -19,6 +19,7 @@ var WalkMe = module.exports = integration('WalkMe')
   .option('trackWalkMeEvents', false)
   .option('loadWalkMeInIframe', false)
   .option('integrityHash', '')
+  .option('bucketName', 'users')
   .tag('<script async="true" src="{{ url }}" crossorigin="" integrity="{{ hash }}">')
 
 /**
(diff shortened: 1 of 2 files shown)


integrations/walkme/lib/index.js
```diff
@@ -19,6 +19,7 @@ var WalkMe = module.exports = integration('WalkMe')
   .option('trackWalkMeEvents', false)
   .option('loadWalkMeInIframe', false)
   .option('integrityHash', '')
+  .option('customDirecotry', 'users')
```
varadarajan-tw, line 22: Is this a typo? Should it be `customDirectory`?

varadarajan-tw (commented):
LGTM! There seems to be a typo in the option name. Depending on the name, I'll create a new option in partner portal for walk me.

integrations/walkme/lib/index.js
paco-walkme: thank you! yeah, this is totally a typo that autocomplete carried over the definition. Thanks for noticing!

varadarajan-tw (approved):

varadarajan-tw:
Hey @paco-walkme, it seems like Custom Directory setting has already been created with the typo and option key is `customDirecotry`. I am unable to delete it. I think we have to revert the typo fix changes and use `customDirecotry` as the option key :( 

image

integrations/walkme/lib/index.js
```diff
@@ -67,7 +68,8 @@ WalkMe.prototype.initialize = function() {
     sriSuffix = 'private_';
   }
 
-  var url = 'https://cdn.walkme.com/users/' + walkMeSystemId + env + '/walkme_' + sriSuffix + walkMeSystemId + '_https.js';
+  var bucket = (this.options.customDirectory) ? this.options.customDirectory : 'users';
+  var url = `https://cdn.walkme.com/${bucket}/${walkMeSystemId}/${env}/walkme_${sriSuffix}${walkMeSystemId}_https.js`;
```
varadarajan-tw, line 72: @paco-walkme - I failed to catch this earlier but after reviewing this PR again my teammate we noticed that you are using template literals here. The integrations in this repo don't support the modern ES6 syntax. Could you change this back to string concatenation with `+`?

varadarajan-tw (changes requested):

integrations/walkme/lib/index.js
paco-walkme: @varadarajan-tw thanks for noticing! I've pushed a fix for it

varadarajan-tw (approved):

varadarajan-tw:
I reverted the key to `customDirecotry`

varadarajan-tw:
Hey @paco-walkme – I will be initiating deployment today. The rollout will be done in three phases over the next three days to reduce risk.
1. Phase 1 - 19 Jul
2. Phase 2- 20 Jul
3. Phase 3 - 21 Jul

If you would like to test this change in Segment production before the full rollout, please share ID of the source connected to your walkme destination in your workspace. I can help deploy this change to your source only for validation.

paco-walkme:
Hello there!
We would love to test the change.

Here is the ID
5b312c44-4264-4eb2-a523-650801cf610e

I don't think i ever done this type of testing in prod, is there a
reference in the documentation about this testing section?

Thanks!

varadarajan-tw:
Np. It is not mandatory to do this testing for the rollout.
If you have a Segment account, you can set up the `walkme` destination just like how customers would set up and test the change ([Walkme docs](https://segment.com/docs/connections/destinations/catalog/walkme/)). I can deploy this change specifically to your instance before rolling out to others so that you can test it and see if it looks okay.

The ID you shared looks doesn't look like a source ID. It would look something like this.
image

PR #785, opened 2023-09-21
[STRATCONN] 2841 pinterest tag mapping event id as segment message id

AnkitSegment:
**What does this PR do?**
Create mapping for pinterest event id as segment messageId

Screenshot 2023-09-21 at 11 37 44 AM

Screenshot 2023-09-21 at 11 52 57 AM

No, 

**Testing**

Testing completed successfully locally

### DIFF
--- a/integrations/pinterest-tag/test/index.test.js
+++ b/integrations/pinterest-tag/test/index.test.js
@@ -94,24 +94,5 @@ describe('Pinterest', function() {
         });
       });
     });
-
-    describe('#track', function() {
-      beforeEach(function() {
-        analytics.spy(window, 'pintrk');
-      });
-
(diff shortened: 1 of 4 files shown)


integrations/pinterest-tag/test/index.test.js
```diff
@@ -9,6 +9,7 @@ var Pinterest = require('../lib/');
 describe('Pinterest', function() {
   var analytics;
   var pinterest;
+  var messageId;
```
varadarajan-tw, line 12: Could you remove the test you wrote for previously? I don't think it is applicable anymore. Is it possible to write a new test that just validations if event_id is set or not? We need not validate the value itself but presence of any value should be fine. This is to ensure that if condition works,

varadarajan-tw (approved):

PR #801, opened 2024-05-30
[MAIN] [STRATCONN-2841] Added event_id in page action of Pinterest Tag

AnkitSegment:
**What does this PR do?**
JIRA TICKET: https://segment.atlassian.net/browse/STRATCONN-2841

Create mapping for Pinterest event_id as segment messageId. 

1. When `mapMessageIdToEventId` is disabled/false

Screenshot 2024-05-30 at 11 57 30 AM

Screenshot 2024-05-30 at 11 56 12 AM

2. When `mapMessageIdToEventId` is enabled/true

Screenshot 2024-05-30 at 11 29 53 AM

Screenshot 2024-05-30 at 11 52 59 AM

**Testing**

- Testing completed successfully using stage 

### DIFF
--- a/integrations/pinterest-tag/lib/index.js
+++ b/integrations/pinterest-tag/lib/index.js
@@ -63,15 +63,27 @@ Pinterest.prototype.identify = function(identify) {
 
 Pinterest.prototype.page = function(page) {
   // If we have a category, the use ViewCategory. Otherwise, use a normal PageVisit.
+  var pinterestPageProps = {
+    name: page.name() || ''
+  };
+
+  var eventKeys = ['event_id', 'eid', 'eventID'];
+
+  for (var i = 0; i < eventKeys.length; i++) {
+    if (page.properties()[eventKeys[i]]) {
+      pinterestPageProps.event_id = page.properties()[eventKeys[i]];
+    }
+  }
+
+  if (this.options.mapMessageIdToEventId) {
+    pinterestPageProps.event_id = page.proxy('messageId');
+  }
+
   if (page.category()) {
-    window.pintrk('track', 'ViewCategory', {
-      category: page.category(),
-      name: page.name() || ''
-    });
+    pinterestPageProps.category = page.category();
+    window.pintrk('track', 'ViewCategory', pinterestPageProps);
   } else {
-    window.pintrk('track', 'PageVisit', {
-      name: page.name() || ''
-    });
+    window.pintrk('track', 'PageVisit', pinterestPageProps);
   }
 };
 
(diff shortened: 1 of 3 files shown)


harsh-joshi99 (approved):
LGTM ✨

integrations/pinterest-tag/lib/index.js
```diff
@@ -63,15 +63,27 @@ Pinterest.prototype.identify = function(identify) {
 
 Pinterest.prototype.page = function(page) {
   // If we have a category, the use ViewCategory. Otherwise, use a normal PageVisit.
+  var pinterestPageProps = {
+    name: page.name() || ''
+  };
+
+  var eventKeys = ['event_id', 'eid', 'eventID'];
```
varadarajan-tw, line 70: Why do we search for `eid`, `eventID` these specific attributes? How did we decide these three values?

varadarajan-tw, line 70:
```suggestion
  const eventKeys = ['event_id', 'eid', 'eventID'];
```

integrations/pinterest-tag/lib/index.js
```diff
@@ -63,15 +63,27 @@ Pinterest.prototype.identify = function(identify) {
 
 Pinterest.prototype.page = function(page) {
   // If we have a category, the use ViewCategory. Otherwise, use a normal PageVisit.
+  var pinterestPageProps = {
```
varadarajan-tw, line 66: you can mutate const object properties
```suggestion
  const pinterestPageProps = {
```

integrations/pinterest-tag/lib/index.js
```diff
@@ -63,15 +63,27 @@ Pinterest.prototype.identify = function(identify) {
 
 Pinterest.prototype.page = function(page) {
   // If we have a category, the use ViewCategory. Otherwise, use a normal PageVisit.
```
varadarajan-tw, line 65: Is this comment more relevant to the `if else` section?

varadarajan-tw (commented):
Looks good. Left a few questions.

Also - could you bump the package version in `package.json`?

integrations/pinterest-tag/lib/index.js
AnkitSegment: [Pinterest only allows event IDs in these three formats. ](https://help.pinterest.com/en/business/article/add-event-codes#:~:text=advertiser%20account%20currency.-,event_id,-eventId0001)

AnkitSegment: In this repo, Eslint is not allowing let and const variable declarations.

varadarajan-tw: Okay. So, do we expect customers to provide these values in properties? Is this consistent with how we can handle `track`?

AnkitSegment: Yes, for `track` events, mappings are generated through the `generatePropertiesObject` function. Using forLoop, when the user adds any of these three event IDs, that value directly gets attached to the mappings.

varadarajan-tw (approved):

PR #818, opened 2025-03-19
Add support to GTM and GA4 for First-Party mode

dja:
**What does this PR do?**
This allows for using Google's First-party mode, to load Google scripts from your own server through server-side tagging.

No

**Testing**

This capability helps ensure that Google Tag Manager domain and script loading is not impeded by browsers, and Segment users can setup the First-party mode to support this.

N/A

- Yes, in the Google Analytics 4 Web and Google Tag Manager destinations, the new setting is `domain`, which defaults to `www.googletagmanager.com`

https://developers.google.com/tag-platform/tag-manager/server-side/dependency-serving?tag=gtag

### DIFF
--- a/integrations/google-analytics-4/HISTORY.md
+++ b/integrations/google-analytics-4/HISTORY.md
@@ -1,3 +1,8 @@
+0.1.0 / 2025-03-19
+==================
+
+  * Add support for custom domain configuration for gtag.js script loading
+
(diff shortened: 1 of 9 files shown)


foreignbyproxy:
Hey there. Super interested in this change. What needs to happen to get this across the line?

